PR #5, opened 2022-06-03
DRAFT - Aw/trainer module

zredlined:
Draft PR for simplifying trainer interfaces. See example in `notebooks/trainer-examples.py`


src/gretel_trainer/trainer.py
```diff
@@ -0,0 +1,156 @@
...
+            round_decimals (int, optional): Round decimals in CSV as preprocessing step. Defaults to 4.
+        """
+        self.dataset_path = dataset_path
+        self.df = self._preprocess_data(
+            dataset_path=dataset_path, round_decimals=round_decimals
+        )
+        self.run = self._initialize_run(df=self.df, overwrite=overwrite)
+        self.run.train_all_partitions()
+
+    def load(self):
```
johntmyers, line 100: Nit: this could probably be a classmethod that actually serves as a constructor. It wasn't obvious to me how to use this and it should probably take the project name as a param since you are using that as part of the manifest name.

src/gretel_trainer/trainer.py
```diff
@@ -0,0 +1,156 @@
...
+
+class Model(namedtuple("Model", "config"), ExtendedEnum):
+    """Enum to pair valid models and configurations"""
+
+    GretelLSTM = "templates/gretel_lstm.yaml"
+    GretelCTGAN = "templates/gretel_ctgan.yaml"
+
+
+class Trainer:
+    """Automated generative training and sampling tool
```
johntmyers, line 44: Nit: orient around "sampling" or "generation" as the term we use but we shouldn't probably mix them

src/gretel_trainer/trainer.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(
+        self,
+        project_name: str = "trainer",
+        max_header_clusters: int = 20,
+        max_rows: int = 50000,
+        model_type: str = "GretelLSTM",
+        model_kwargs: dict = {}
+    ):
+
+        configure_session(api_key="prompt", cache="yes", validate=True)
```
johntmyers, line 63: Issue: if I already have a client configured with this force me to re-authenticate? We may want to first check if the client is already configured and able to make api calls first before forcing a re-auth

src/gretel_trainer/trainer.py
```diff
@@ -0,0 +1,156 @@
...
+
+    @classmethod
+    def get_config(cls, model: str):
+        return cls[model].config
+
+
+class Model(namedtuple("Model", "config"), ExtendedEnum):
+    """Enum to pair valid models and configurations"""
+
+    GretelLSTM = "templates/gretel_lstm.yaml"
```
johntmyers, line 39: Qq: any reason not to use the default configs we have on the blueprints repo and just access them by their namespace like you can now with the client? This feels like we are replicating default configs across different repos. Are these configs more optimized for trainer usage?

src/gretel_trainer/trainer.py
```diff
@@ -0,0 +1,156 @@
...
+    def _preprocess_data(
+        self, dataset_path: str, round_decimals: int = 4
+    ) -> pd.DataFrame:
+        """Preprocess input data"""
+        tmp = pd.read_csv(dataset_path, low_memory=False)
+        tmp = tmp.round(round_decimals)
+        return tmp
+
+    def _initialize_run(
+        self, df: pd.DataFrame = pd.DataFrame(), overwrite: bool = True
```
johntmyers, line 130: Nit: set the default df value to None and them create an empty dataframe in the actual function, this could have some side effects on the method itself if you have mutable objects as default params

src/gretel_trainer/trainer.py
zredlined: If you already have a client configured it will use your existing credentials and will **not** prompt you to re-auth.

src/gretel_trainer/trainer.py
zredlined: Modified the trainer to use configs loaded from the blueprints repo

johntmyers (approved):

PR #12, opened 2022-08-04
[DRAFT] Aw/conditional generation

zredlined:
* Added support for CTGAN
* Model specific configs now supported
* Conditional data generation with LSTM and CTGAN models
* Examples in python and ipynb formats


src/gretel_trainer/models.py
```diff
@@ -0,0 +1,81 @@
...
+    @classmethod
+    def get_default_header_clusters(cls, model: str) -> int:
+        return cls[model].config.default_header_clusters
+
+    @classmethod
+    def get_max_header_clusters(cls, model: str) -> int:
+        return cls[model].config.max_header_clusters
+
+
+class Model(namedtuple("Model", "config"), ExtendedEnum):
```
johntmyers, line 77: I'm not sure I understand the purpose of this enum as well as the wrapper around it with the extended enum above. Couldn't we just use instances of `BaseConfig` as a param to the `Trainer` class?

Instead of `model_type: str` as a param to `Trainer` could we just pass in a `model_config: BaseConfig`? This is similar to what we do in `gretel-synthetics` and just use a builder pattern.

Also on the `Trainer` I'm not sure why `max_rows` and `max_header_clusters` need to params, it seems like they could just be only defined on the config instance?

So trainer could be something like:

`trainer = Trainer(model_config=GretelLSTMConfig()` then you don't need all the getters from the enum you can just access the attributes directly like `trainer.model_config.max_rows`

src/gretel_trainer/trainer.py
```diff
@@ -58,43 +40,56 @@ class Trainer:
...
         self.project = create_or_get_unique_project(name=project_name)
-        self.max_header_clusters = max_header_clusters
-        self.max_rows = max_rows
         self.overwrite = overwrite
         self.cache_file = self._get_cache_file(cache_file)
 
-        if model_type in Model.get_types():
-            self.config = read_model_config(Model.get_config(model_type))
+        # Verify model specific settings
+        if model_type not in Model.get_types():
```
johntmyers, line 64: If we used a `model_config: BaseConfig` here, you could just do an `isinstance(model_config, BaseConfig)` to ensure we have a valid config

src/gretel_trainer/trainer.py
```diff
@@ -58,43 +40,56 @@ class Trainer:
...
+            raise ValueError(
+                f"Invalid model type specified. Must be one of {Model.get_types()}."
+            )
+        if self.max_rows and self.max_rows >= Model.get_max_rows(model_type):
```
johntmyers, line 68: You could enforce these values directly on the config instances themselves (see above)

setup.py
```diff
@@ -5,7 +5,7 @@
 install_requires = (local_path / "requirements.txt").read_text().splitlines()
 
 setup(name="gretel-trainer",
-      version="0.2.1",
+      version="0.2.2",
```
johntmyers, line 8: we should change this to pull from github tags instead of manually doing it - we can set that up in a separate pr

src/gretel_trainer/models.py
johntmyers: I didn't test this, but this would give you a more self-contained config that is self-validating. Not sure we need to swap to something like `pydantic` just yet. You can define your constraints i.e. `_max_rows` as actual _class_ attributes because numbers and strings are immutable. This way you don't have to clutter up the constructors with constraints the user shouldn't really need to think about.

Also, on the concrete constructors, you will probably want to have your default values declared on the actual `__init__` because as it stands now there's no way to change these values. It looks like you were actually trying to replicate these values into the `Trainer` as attributes, but I'd suggest just keeping everything contained within the config class instance.

```python
class _BaseConfig:
    """This class should not be used directly, model-specific configs should be
    derived from this class
    """

    _max_header_clusters: int
    """This should be overridden on conrete classes"""

    # Default values for all configs, may be overridden if needed
    _max_rows: int = 1000000

    # Should be set by concrete constructors
    config_file: str
    default_rows: int
    default_header_clusters: int

    def __init__(
        self,
        config_file: str,
        default_rows: int,
        default_header_clusters: int,
    ):
        self.config_file = config_file
        self.default_rows = default_rows
        self.default_header_clusters = default_header_clusters

        self.validate()

    def validate(self):
        if self.default_rows > self._max_rows:
            raise ValueError("too many rows")

        if self.default_header_clusters > self._max_header_clusters:
            raise ValueError("too many header clusters")


class GretelLSTMConfig(_BaseConfig):

    _max_header_clusters: int = 30

    def __init__(
        self,
        config_file="synthetics/default",
        default_rows=50000,
        default_header_clusters=20,
    ):
        super().__init__(
            config_file=config_file,
            default_rows=default_rows,
            default_header_clusters=default_header_clusters,
        )


class GretelCTGANConfig(_BaseConfig):

    _max_header_clusters: int = 1000

    def __init__(
        self,
        config_file="synthetics/high-dimensionality",
        default_rows=50000,
        default_header_clusters=500,
    ):
        super().__init__(
            config_file=config_file,
            default_rows=default_rows,
            default_header_clusters=default_header_clusters,
        )
```

zredlined: I used the Enum to let the code enumerate the available model types and check that whatever the user was asking for was a valid type. But agreed it's a little extra and we can add this to the class functions anyways. Let's go with the changes above 👍

setup.py
zredlined: that's good, never liked doing that in code 👍

johntmyers, line 8: can we drop this change?

src/gretel_trainer/models.py
```diff
@@ -0,0 +1,107 @@
...
+        self.max_header_clusters = max_header_clusters
+        self.enable_privacy_filters = enable_privacy_filters
+
+        if not self.enable_privacy_filters:
+            logging.warning("Privacy filters disabled. Enable with the `enable_privacy_filters` param.")
+            self.update_params({"outliers": None, "similarity": None})
+
+        self.validate()
+
+    def update_params(self, params:dict):
```
johntmyers, line 42: nit: `params: dict` (missing a whitespace) - also a future improvement here would to make sure that the config we're operating on is "fully qualified", if someone provided a config that didn't have every possible field on it, updating params may have no effect, is that right?

johntmyers (approved):
One nit and curious about how to make the param replacement more stable for partial configs

src/gretel_trainer/models.py
```diff
@@ -0,0 +1,107 @@
...
+
+    _max_header_clusters: int
+    """This should be overridden on concrete classes"""
+
+    # Default values for all configs, may be overridden if needed
+    _max_rows: int = 1000000
+    _enable_privacy_filters: bool = False
+
+    # Should be set by concrete constructors
+    config: Union[str, dict]
```
pimlock, line 20: Nit: `self.config` is always `dict`.
```suggestion
    config: dict
```

src/gretel_trainer/models.py
```diff
@@ -0,0 +1,107 @@
...
+    """This class should not be used directly, models should be
+    derived from this class
+    """
+
+    _max_header_clusters: int
+    """This should be overridden on concrete classes"""
+
+    # Default values for all configs, may be overridden if needed
+    _max_rows: int = 1000000
+    _enable_privacy_filters: bool = False
```
pimlock, line 17: This is not used. Should it be used somewhere? Or can it be deleted?
Or maybe it shouldn't have the `_`?

src/gretel_trainer/models.py
```diff
@@ -0,0 +1,107 @@
...
+class _BaseConfig:
+    """This class should not be used directly, models should be
+    derived from this class
+    """
+
+    _max_header_clusters: int
+    """This should be overridden on concrete classes"""
+
+    # Default values for all configs, may be overridden if needed
+    _max_rows: int = 1000000
```
pimlock, line 16: Nit: I think naming these fields something like `_max_header_clusters_limit` and `_max_rows_limit` would make it a bit more readable, for now it's very similar to `max_header_clusters` and it's hard to tell why one has `_` and the other doesn't.

pimlock (approved):
Looks good, just few small things.

src/gretel_trainer/models.py
zredlined: This is a good call. I'll create an issue for the next release. We should probably add some logic to validate that the passed-in config is valid for the model being called. It should be pretty simple- we could either validate all config options (more work), or just verify that the correct model slug (e.g. synthetics, ctgan, etc) is used in the config

src/gretel_trainer/models.py
zredlined: In the constructor, a config can be a string- e.g. "synthetics/default" or a dictionary

src/gretel_trainer/models.py
zredlined: These can be deleted

src/gretel_trainer/models.py
zredlined: Added a simple check to confirm that the model slug matches the selected model_type. Will build a more advanced check to account for use cases above in a future release.

src/gretel_trainer/models.py
pimlock: It can be a string only in the `__init__`, but this field will always be a `dict`, since it's initialized as `self.config = read_model_config(config)` ([link](https://github.com/gretelai/trainer/pull/12/files#diff-2319392ddaacefb2b7073a640aaa686428a65a968dc11262c1ad5ae94ae9d407R29)).

In the `__init__`, it's correctly annotated as `config: Union[str, dict]`.

PR #13, opened 2022-08-04
Add param to pass custom base configuration

mikeknep:
- Prefer `config` if present, otherwise use the `model_type`'s default config.
- This does open the door a little wider to setting an invalid config that won't be known to be bad until attempting to train. That door was already slightly ajar in that one could use `model_params` to set keys to invalid values.
- Not included here, but a thought: we could validate `model_type` earlier (even as the very first step of `__init__`) to fail fast, specifically before even creating a project.


mikeknep:
Closing this as Alex is working the same general change into his PR.

PR #14, opened 2022-08-08
Get average sqs score from across partitions

mikeknep:
A few ways we could slice and dice this; I figure there may be additional SQS info we want from the run in the future so I decided to expose the entire `List[dict]` from the runner, and let the trainer pluck out and calculate the first such aggregate, user-friendly data. I'm open to pushing more of this down to the runner and/or transforming the SQS dictionaries into first-class types (likely dataclasses) if anyone has a strong opinion or thinks it'd be useful.

### DIFF
--- a/src/gretel_trainer/trainer.py
+++ b/src/gretel_trainer/trainer.py
@@ -112,7 +112,7 @@ class Trainer:
         )
         return self.run.get_synthetic_data()
 
-    def get_sqs_score(self) -> float:
+    def get_sqs_score(self) -> int:
         """Return the average SQS synthetic data quality score.
 
         Requires the model has been trained.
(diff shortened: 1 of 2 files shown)


src/gretel_trainer/trainer.py
```diff
@@ -151,6 +151,17 @@ def generate(self, num_records: int = 500) -> pd.DataFrame:
...
+    def get_sqs_score(self) -> float:
+        """Return the average SQS synthetic data quality score.
+
+        Requires the model has been trained.
+        """
+        scores = [
+            sqs["synthetic_data_quality_score"]["score"]
+            for sqs in self.run.get_sqs_information
+        ]
+        return sum(scores) / len(scores)
```
zredlined, line 124: We should probably cast this return to `int`. Our SQS scores are always integers, this could return a float

zredlined (approved):

PR #15, opened 2022-08-16
Auto-determine best model from training data

mikeknep:
Rather than create a `GretelAuto` model class that would need to override or work around several `_BaseConfig` details (validation, max/limit values, etc.), my goal here is to establish the convention that model type is optional and if you don't specify one when instantiating the Trainer, you're OK with us choosing for you. This is a change from the current behavior (optional but default to LSTM). In this case, we defer setting the trainer instance's `self.model_type` until such time as we can determine the best model to use: namely, at train time when a dataset has been provided.

I'm a little unclear on the `load` (from cache) workflow, which in this branch's implementation would set the StrategyRunner's `model_config` to `None`. I think this is OK because the only methods referencing that value are part of training (`train_all_partitions` => `train_next_partition` => `train_partition`), and that workflow is only kicked off by the Trainer's `train` method, which will load in data and use it to determine and set a concrete model.

I've also added an optional `delimiter` parameter to `train` to help support files with non-comma delimiters.

### DIFF
--- a/src/gretel_trainer/models.py
+++ b/src/gretel_trainer/models.py
@@ -1,7 +1,26 @@
 import logging
 from typing import Union
 
+import pandas as pd
+
 from gretel_client.projects.models import read_model_config
+
+
+HIGH_COLUMN_THRESHOLD = 20
+HIGH_RECORD_THRESHOLD = 50000
+LOW_COLUMN_THRESHOLD = 4
+LOW_RECORD_THRESHOLD = 1000
+
+
+def determine_best_model(df: pd.DataFrame):
+    row_count, column_count = df.shape
+
+    if row_count > HIGH_RECORD_THRESHOLD or column_count > HIGH_COLUMN_THRESHOLD:
+        return GretelCTGAN(config="synthetics/high-dimensionality")
+    elif row_count < LOW_RECORD_THRESHOLD or column_count < LOW_COLUMN_THRESHOLD:
+        return GretelCTGAN(config="synthetics/low-record-count")
+    else:
+        return GretelLSTM()
 
 
 class _BaseConfig:
--- a/src/gretel_trainer/models.py
+++ b/src/gretel_trainer/models.py
@@ -23,7 +23,7 @@ def determine_best_model(df: pd.DataFrame):
     elif row_count < LOW_RECORD_THRESHOLD or column_count < LOW_COLUMN_THRESHOLD:
         return GretelCTGAN(config="synthetics/low-record-count")
     else:
-        return GretelLSTM()
+        return GretelLSTM(config="synthetics/default")
 
 
 class _BaseConfig:
--- a/src/gretel_trainer/models.py
+++ b/src/gretel_trainer/models.py
@@ -1,7 +1,10 @@
-import logging
-from typing import Union
+from __future__ import annotations
 
-import pandas as pd
+import logging
+from typing import TYPE_CHECKING, Union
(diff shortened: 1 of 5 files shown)


src/gretel_trainer/models.py
```diff
@@ -1,9 +1,28 @@
 import logging
 from typing import Union
 
+import pandas as pd
```
mikeknep, line 4: This is imported only for the type hint in `determine_best_model`... I'm open to the argument that this isn't necessary and we just don't type hint that function

src/gretel_trainer/trainer.py
```diff
@@ -47,7 +47,7 @@ def __init__(
         self.cache_file = self._get_cache_file(cache_file)
         self.model_type = model_type
 
-        if self.overwrite:
+        if self.overwrite and self.model_type is not None:
             logger.debug(json.dumps(self.model_type.config, indent=2))
```
mikeknep, line 51: Would it be worth logging something about "running with auto-config" in the case where `self.model_type is None`, or fine as-is?

src/gretel_trainer/models.py
pimlock: We've been using this pattern in cases where import is only used for type hinting and I think that's what would work well here as well.

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import pandas as pd
```

src/gretel_trainer/models.py
```diff
@@ -1,9 +1,31 @@
...
+
+def determine_best_model(df: pd.DataFrame):
+    row_count, column_count = df.shape
+
+    if row_count > HIGH_RECORD_THRESHOLD or column_count > HIGH_COLUMN_THRESHOLD:
+        return GretelCTGAN(config="synthetics/high-dimensionality")
+    elif row_count < LOW_RECORD_THRESHOLD or column_count < LOW_COLUMN_THRESHOLD:
+        return GretelCTGAN(config="synthetics/low-record-count")
+    else:
+        return GretelLSTM()
```
zredlined, line 26: We should pass in a config here as well explicitly. E.g. config="synthetics/default"

src/gretel_trainer/trainer.py
zredlined: Yes, something to the effect of "Automatically selecting optimal model config" would be great.

zredlined (changes requested):
Currently the trainer generates a dataset the same size as the input when the `train()` function is run. Let's update the default behavior to match our SDK, generating 5k records (and an SQS score) per partition. This should make training faster and more flexible as the user can specify how many records they want in the `generate()` function

src/gretel_trainer/runner.py
```diff
@@ -413,11 +414,6 @@ def train_partition(
 
         if "synthetics" in model_config["models"][0].keys():
 
-            model_config["models"][0]["synthetics"]["generate"] = {
-                "num_records": artifact.record_count,
-                "max_invalid": None,
-            }
```
mikeknep, line 419: @zredlined Removing this restores the behavior of generating 5000 records during training, which is both [explicitly on the `synthetics/default` blueprint](https://github.com/gretelai/gretel-blueprints/blob/main/config_templates/gretel/synthetics/default.yml#L33) and is the default when unspecified, as in the case of CTGAN (both here in the trainer and in its blueprint)

zredlined (approved):

PR #20, opened 2022-10-05
Use artifact manifest for determine_best_model.

kboyd:
Not fully tested. Waiting for new backend API to be available.

Should revisit retry logic if we can reliably distinguish between a pending manifest (still being generated) and some other error. Or if retrying is included in the `gretel_client` interface.


src/gretel_trainer/trainer.py
```diff
@@ -193,3 +195,22 @@ def _initialize_run(
...
+            target_file = str(Path(tmp) / "trainer-original.csv")
+            df.to_csv(target_file, index=False)
+            artifact_id = self._project.upload_artifact(target_file)
+
+        return self._fetch_artifact_manifest(artifact_id)
+
+    @retry(tries=5, delay=1, backoff=2)
+    def _fetch_artifact_manifest(self, artifact_id: str) -> dict:
+        """Request (with retries) the manifest for an existing artifact_id."""
+        manifest = project.get_artifact_manifest(artifact_id)
```
johntmyers, line 214: Should this be `self.project?`

src/gretel_trainer/trainer.py
```diff
@@ -193,3 +195,22 @@ def _initialize_run(
...
+
+    def _generate_artifact_manifest(self, df: pd.DataFrame) -> dict:
+        """Upload DataFrame to API to generate manifest.
+
+        Blocks until manifest is generated and returned.
+        """
+        with tempfile.TemporaryDirectory() as tmp:
+            target_file = str(Path(tmp) / "trainer-original.csv")
+            df.to_csv(target_file, index=False)
+            artifact_id = self._project.upload_artifact(target_file)
```
johntmyers, line 207: `self.project`?

src/gretel_trainer/trainer.py
kboyd: Fixed.

src/gretel_trainer/trainer.py
kboyd: Fixed.

mikeknep:
@kboyd are you still interested in landing these changes? I did a quick skim of the diff and there's some outdated stuff by now (ex. we already have a `GretelAmplify` model as of November). I can dive deeper into the PR if we still want to ultimately land it, though it would help if you could merge/rebase in the latest from `main` first.

kboyd:
Thanks for the reminder. We switched to focus on auto params within each model and will revisit auto model type selection in the future, but we don't need this PR in the meantime and doesn't make sense to try and merge it if we aren't actively using the functionality.

Closing this draft PR so it doesn't clutter everything up.

PR #25, opened 2022-10-13
Requirements.txt fixed for Sphinx docs build

MasonEgger:
Getting dependencies installed was a chore, but I got it working. Just need to merge in.


MasonEgger:
This isn't ready, I found an error. Will submit again when ready.

PR #28, opened 2022-10-21
adding model docs

MasonEgger:


src/gretel_trainer/models.py
```diff
@@ -16,6 +16,16 @@
 
 
 def determine_best_model(df: pd.DataFrame):
+    """
+    Determine which Gretel model would best suited for generating synthetic data
```
csbailey5t, line 20:
```suggestion
    Determine the Gretel model best suited for generating synthetic data
```

src/gretel_trainer/models.py
```diff
@@ -16,6 +16,16 @@
...
+    for your dataset. 
+
+    Args:
+        df (pd.DataFrame): Pandas DataFrame containing the data you intend to synthesize.
```
csbailey5t, line 24: I think this should be something like "the data used to train a synthetic model"

src/gretel_trainer/models.py
```diff
@@ -16,6 +16,16 @@
...
+
+    Returns:
+        A Gretel Model object preconfigured suited for your use case.
```
csbailey5t, line 27:
```suggestion
        A Gretel Model object preconfigured for your use case.
```

src/gretel_trainer/models.py
```diff
@@ -93,7 +103,17 @@ def _replace_nested_key(self, data, key, value) -> dict:
 
 
 class GretelLSTM(_BaseConfig):
+    """
+    This model works for a variety of synthetic data tasks including time-series, tabular, and text data. Generally useful for a few thousand records and upward. Dataset generally has a mix of categorical, continuous, and numerical values
 
+    Source data should have <150 columns.
+
+    Args:
+        config (str/dict, optional): Either a string representing the path on to the config on the local filesystem, a string representing a path to the default Gretel configurations, or a dictionary containing the configurations. Default: "synthetics/default", a default Gretel configuration
```
csbailey5t, line 112:
```suggestion
        config (str/dict, optional): Either a string representing the path to the config on the local filesystem, a string representing a path to the default Gretel configurations, or a dictionary containing the configurations. Default: "synthetics/default", a default Gretel configuration
```

csbailey5t (dismissed):
A few minor copy changes

src/gretel_trainer/models.py
MasonEgger: Updated in a new commit

csbailey5t (approved):

PR #30, opened 2022-10-26
Lift gretel model compatibility to separate module

mikeknep:
### What's here

Make it easier to find the "compatibility rules" for models by lifting the logic to its own module.

#### Why not add this logic to the specific model classes? Wouldn't that be more polymorphic?

The model classes (`GretelLSTM`, `GretelCTGAN`, etc.) are wrappers around _specific configurations_ from the blueprints repo. They do not represent every possible configuration of that model type. If a user wants to run a customized LSTM config, for example, they subclass `GretelModel`, _not_ `GretelLSTM`:
```python
class MyLstm(GretelModel):
    config = "/path/to/my_lstm.yml"
```
Note: they _could_ subclass `GretelLSTM`, but 1) it's easier to tell people to just subclass `GretelModel` regardless of model type, and/because 2) this ultimately treats the model configuration as the source of truth.

If someone mistakenly created a custom Gretel model like this...
```python
class MyGptX(GretelGPTX):
    config = "/path/to/my_amplify.yml"
```
...Benchmark will treat this as an Amplify model, because basically all it does with the class instance is grab the config attribute (and the name—the results output will show the name as `MyGptX`.)

### DIFF
--- a/src/gretel_trainer/benchmark/gretel/compatibility.py
+++ b/src/gretel_trainer/benchmark/gretel/compatibility.py
@@ -1,0 +1,34 @@
+from typing import Optional
+
+from gretel_trainer.benchmark.core import DataSource, Datatype
+
+
+def is_runnable(model_key: Optional[str], source: DataSource) -> bool:
+    if model_key is None:
+        return True
+    elif model_key in ("lstm", "synthetics"):
+        return _lstm(source)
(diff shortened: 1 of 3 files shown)


mikeknep:
Closing this in favor of a larger refactor of benchmark models, which will be in a separate PR.

PR #31, opened 2022-11-02
Configure session before starting Benchmark comparison

mikeknep:
### Current behavior

When running in an environment where no Gretel credentials can be found (e.g. Colab), when Benchmark kicks off a comparison the background threads instantiating Trainer instances will prompt for an API key. This is problematic for multiple reasons, all (I believe) due to it running in multiple background threads: it prompts multiple times, doesn't accept input and/or cache properly, and ultimately crashes.

### This fix

Benchmark itself now checks for a configured session before kicking off any real work. It prompts (`api_key="prompt"`) if no credentials are found, validates (`validate=True`) the supplied API key, and caches (`cache="yes"`) it for all the runs it manages. The `configure_session` calls that happen when instantiating Trainer effectively "pass through." I've tested this by installing trainer from this branch in Colab and it is now working as expected.

### DIFF
--- a/src/gretel_trainer/benchmark/compare.py
+++ b/src/gretel_trainer/benchmark/compare.py
@@ -157,6 +157,8 @@ def compare(
     gretel_sdk: GretelSDK,
     gretel_trainer_factory: Callable[..., Trainer],
 ) -> Comparison:
+    gretel_sdk.configure_session()
+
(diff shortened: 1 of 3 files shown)


pimlock (approved):
Nice! This makes sense, thanks or the detailed explanation in description.

PR #33, opened 2022-11-08
Switch CTGAN usages to ACTGAN.

pimlock:
ACTGAN is the successor of CTGAN.

Note (1): this change is backward compatible, as all of the parameters that CTGAN supported are supported by ACTGAN as well.

Note (2): any previously trained CTGAN models will be still usable, i.e. it will be possible to generate new records using old CTGAN models.


src/gretel_trainer/models.py
```diff
@@ -145,27 +143,42 @@ class GretelCTGAN(_BaseConfig):
...
 
     def __init__(
         self,
         config="synthetics/high-dimensionality",
-        max_rows=50_000,
-        max_header_clusters=500,
-        enable_privacy_filters=False,
+        max_rows=1_000_000,
+        max_header_clusters=1_000,
+        enable_privacy_filters=None,
```
mikeknep, line 158: Is it worth adding a log warning about this? Something like:
```python
if enable_privacy_filters is not None:
    logger.warning("enable_privacy_filters is deprecated...")
```

mikeknep (approved):
Thanks for picking up all these!

src/gretel_trainer/models.py
pimlock: Yeah, that's a good idea!

PR #34, opened 2022-11-10
Benchmark route Amplify models through Trainer

mikeknep:
### Top level change
Now that Trainer has a GretelAmplify model, Benchmark uses Trainer for Amplify runs instead of the SDK.

### Refactor
I refactored Benchmark's Gretel models and executors with the goal of centralizing and thus making it simpler to understand:
- which model types use Trainer (opt-in) vs. use the SDK
- the "compatibility requirements" for different models (currently: LSTM <= 150 columns, GPTX == 1 column)

These had been spread across a few different places (`compare.py` determined Trainer/SDK, `gretel/sdk.py` had GPTX compatibility, `gretel/trainer.py` had LSTM compatibility), but now it can all be found in `gretel/models.py`.

At first glance it would seem compatibility requirements could be defined on specific model subclasses to make things more polymorphic. However, Benchmark's Gretel model classes are really just friendly wrappers around _specific model configurations_ (from the blueprints repo) and do not represent all possible instances of that model type running through Benchmark. Instead, we [instruct users](https://docs.gretel.ai/gretel.ai/evaluation-framework/gretel-benchmark#docs-internal-guid-0541ad86-7fff-c59f-d506-94d5782f482f) subclass the generic `GretelModel` base class when they want to provide their own specific Gretel configuration. There are two reasons for this:
1. It's a simpler instruction (always subclass this one thing)
2. It enables us to include model types that are not yet "first class supported," such as DGAN (which we can't support in the same way we do models like Amplify/LSTM/etc. because DGAN's config includes required fields that are specifically coupled to the data source—there is no "one size fits all" blueprint).

### Small fixes

- fix the `model_slug` value for Trainer's `GretelACTGAN` model 
  - :warning: should this be changed to a list `["actgan", "ctgan"]` for a little while for a smoother transition/deprecation experience??
- zero-index custom model runs' run-identifier to match gretel model runs (which were themselves fixed to match project names [here](https://github.com/gretelai/trainer/pull/32))

### DIFF
--- a/src/gretel_trainer/benchmark/gretel/executor.py
+++ b/src/gretel_trainer/benchmark/gretel/executor.py
@@ -1,0 +1,56 @@
+from typing import Callable, Union
+
+import pandas as pd
+
+from gretel_trainer.benchmark.core import DataSource
+from gretel_trainer.benchmark.gretel.models import GretelModel
+from gretel_trainer.benchmark.gretel.sdk import GretelSDK, GretelSDKExecutor
+from gretel_trainer.benchmark.gretel.trainer import GretelTrainerExecutor, Trainer
+
+
+class GretelExecutor:
+    def __init__(
+        self,
+        project_name: str,
+        model: GretelModel,
+        sdk: GretelSDK,
+        trainer_factory: Callable[..., Trainer],
+        benchmark_dir: str,
+    ):
+        self._project_name = project_name
+        self._model = model
+        self._internal_executor = _choose_executor(project_name, model, sdk, trainer_factory, benchmark_dir)
+
+    @property
+    def model_name(self) -> str:
+        return self._model.name
+
+    def runnable(self, source: DataSource) -> bool:
+        return self._model.runnable(source)
+
+    def train(self, source: str, **kwargs) -> None:
+        self._internal_executor.train(source, **kwargs)
+
+    def generate(self, **kwargs) -> pd.DataFrame:
+        return self._internal_executor.generate(**kwargs)
+
+    def get_sqs_score(self, synthetic: pd.DataFrame, reference: str) -> int:
+        return self._internal_executor.get_sqs_score(synthetic, reference)
+
+
+def _choose_executor(
+    project_name: str,
+    model: GretelModel,
+    sdk: GretelSDK,
+    trainer_factory: Callable[..., Trainer],
+    benchmark_dir: str,
+) -> Union[GretelTrainerExecutor, GretelSDKExecutor]:
+    if model.use_trainer:
+        return GretelTrainerExecutor(
+            project_name,
+            model.trainer_model_type,
+            trainer_factory,
+            benchmark_dir
+        )
+    else:
+        return GretelSDKExecutor(project_name, model, sdk)
--- a/src/gretel_trainer/benchmark/gretel/models.py
+++ b/src/gretel_trainer/benchmark/gretel/models.py
@@ -1,15 +1,34 @@
 from pathlib import Path
-from typing import Dict, Union
+from typing import Dict, Optional, Tuple, Union
 
+import gretel_trainer
+
+from gretel_client.projects.exceptions import ModelConfigError
 from gretel_client.projects.models import read_model_config
+from gretel_trainer import models
+from gretel_trainer.benchmark.core import BenchmarkException, DataSource, Datatype
 
 GretelModelConfig = Union[str, Path, Dict]
 
 
-def _config(blueprint: str, class_name: str) -> Dict:
-    config = read_model_config(blueprint)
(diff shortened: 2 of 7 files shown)


src/gretel_trainer/models.py
```diff
@@ -148,7 +148,7 @@ class GretelACTGAN(_BaseConfig):
 
     _max_header_clusters_limit: int = 5_000
     _max_rows_limit: int = 5_000_000
-    _model_slug: str = "ctgan"
+    _model_slug: str = "actgan"
```
pimlock, line 151: Good catch, thanks for fixing this!

src/gretel_trainer/benchmark/gretel/models.py
```diff
@@ -1,15 +1,34 @@
...
+    "amplify": models.GretelAmplify,
+    "ctgan": models.GretelACTGAN,
+    "synthetics": models.GretelLSTM,
+
+    # Benchmark GretelAuto sends None model_type to Trainer to trigger dynamic model selection
+    "AUTO": lambda _: None,
+}
+
+
+def _is_compatible(model_key: Optional[str], source: DataSource) -> bool:
```
pimlock, line 25: With the current usage of this, it looks like `model_key` will be a `str`, since for auto mode it returns `AUTO`.

src/gretel_trainer/benchmark/gretel/executor.py
```diff
@@ -0,0 +1,56 @@
...
+        return self._internal_executor.get_sqs_score(synthetic, reference)
+
+
+def _choose_executor(
+    project_name: str,
+    model: GretelModel,
+    sdk: GretelSDK,
+    trainer_factory: Callable[..., Trainer],
+    benchmark_dir: str,
+) -> Union[GretelTrainerExecutor, GretelSDKExecutor]:
```
pimlock, line 47: Could this return be hinted just as `Trainer`? Which is the protocol that you defined.

pimlock (approved):
Looks great!
Just few minor comments.

pimlock:
> * ⚠️ should this be changed to a list `["actgan", "ctgan"]` for a little while for a smoother transition/deprecation experience??

This is a great question.
Theoretically, with your change, the `GretelCTGAN` class should have `_slug` set to `ctgan`? If someone has old config and we make a decision based on that config before our API is called, then we need to keep `ctgan` there?

I think in general, handling aliasing outside of our API isn't working great, because on one hand we wanted the transition to be transparent to the user (i.e. replace the name in the API, so no matter how users are using us, they will get the same experience), but on the other hand we have these constructs in trainer/client that should be updated, while still preserving backward compatibility, so no matter if you are using new client or old client, your code or CLI commands should work.
It's quite tricky to get it right with how we do aliasing, but maybe we are in a good enough state?

src/gretel_trainer/benchmark/gretel/executor.py
mikeknep: This is a really interesting question.

I don't think hinting the return value as `Trainer` is appropriate here because that protocol is designed to capture the contract Benchmark has with the actual `gretel_trainer.Trainer` class. The fact that it is currently the exact same interface is (somewhat) incidental, and more importantly may not always be the case. For example, as Evaluate evolves to provide more info about synthetic datasets, we may want the "internal executors" (`GretelTrainerExecutor` and `GretelSDKExecutor`) to expose a single method—something like `get_synthetic_evaluation`—that under the hood may may require the GretelTrainerExecutor to aggregate results of a few Trainer methods (maybe something like `get_sqs_score` + `get_some_other_info` + `get_a_third_aspect`). I would add those low-level operations to the `Trainer` protocol to indicate "here's what I-the-client-Benchmark expect to be able to do with Trainer-the-library" / "these are the public functions from Trainer that I'm using". (Also worth noting, the mechanism for fetching that same enhanced Evaluate data from the SDK would likely be a little different, and so that branch of Gretel model execution would not be able to adhere to the same interface expected by this return type if it were `Trainer`.)

This question did make me wonder if there is some other protocol we could use here instead of the concrete union type, but ultimately I think the current return type hint is most appropriate. It expresses the fact that we're choosing between one of two (and only ever two) options, and those two options are really just private implementation details of the `GretelExecutor` class. To that end, I'm pushing a commit that makes that even more clear by bringing those classes into the same module. I think it actually makes the purposes of _three_ modules (executor, trainer, and sdk) a bit more clear—let me know if you agree!

pimlock: I see, thank you for all the context on this!
> It expresses the fact that we're choosing between one of two (and only ever two) options, and those two options are really just private implementation details of the `GretelExecutor` class.

This makes perfect sense, thanks!

pimlock (approved):
Nicely done!💯

PR #39, opened 2023-01-23
Hello MultiTable!

mikeknep:
Adds v0 of relational data support under `gretel_trainer.relational`. Since that's all greenfield code, I'll summarize the other changes outside of that scope present in this PR.

- Adds CI action to run type-checking and unit tests. (Note: trainer core is omitted from type-check scope for now; would be nice to add it in the future but there are several warnings to address.)
- Some helper Make targets for use while working on relational stuff
- Simplify how `Trainer` class is imported, and update notebook examples accordingly
- Adds a few dependencies required by relational. Most of these are added to the main `requirements.txt` file, but we also add a few additional deps under the `[relational]` "extras" flag to install specific database drivers that can be used with relational's `Connector`


gracecvking (approved):
LGTM! Let's ship

zredlined (approved):

PR #40, opened 2023-01-25
Dupe logs

mikeknep:
It appears Colab preconfigures a default root handler that, when combined with our customized log setup, leads to duplicate log output. This change ensures any root handlers are cleared out before we set up our custom ones. Appears to be working as expected now, and local behavior appears unchanged.


gracecvking (approved):
Impressively fast fix, Mike! LGTM

PR #41, opened 2023-01-25
Rename strategies and adjust defaults

mikeknep:
- Renames strategies
  - single-table => independent
  - cross-table => ancestral
- Adds deprecation warnings when previous strategy config values are provided
- Use Independent strategy by default
- Continue defaulting to Amplify regardless of strategy


gracecvking (approved):
LGTM. Tested strategy="single-table" and strategy="cross-table" to ensure that it still works, and shows depreciation message.

PR #42, opened 2023-01-26
Rename and split connector extras

mikeknep:
The previous name "relational" was a bit misleading, since you don't need to install those extras to use the relational module generally; they are only required for various connectors.

This change gives a clearer name to that extras package, and also splits by db type to allow finer-grained package installation if desired.


gracecvking (approved):
👍

PR #44, opened 2023-01-26
Seed performance

mikeknep:
Builds a list of parent indices and merges parent data all at once instead of extracting and appending entire records in each iteration of a large loop.

Also drops `itertools.cycle` (from both strategies) and "reverts" back to classic index-incrementing given `cycle`'s [warning about auxiliary storage requirements](https://docs.python.org/3/library/itertools.html#itertools.cycle).


amysteier (approved):
looks great

gracecvking (approved):

PR #51, opened 2023-02-03
Keyword args

mikeknep:
Requires you specify argument names, rather than pass values positionally. See notebook diff for examples.

Also renames the notebook from multitable to relational.


gracecvking (approved):
GTG. I added one commit adjusting the title header within the relational notebook. Made it Gretel Relational instead of Multitable

PR #52, opened 2023-02-03
Upload source on init

mikeknep:
Instead of waiting to upload source CSVs to the project until calling `train`, with this change we will upload them when the MultiTable instance is created. Two reasons for this:
1. Allows restoring from just a created MT instance that hasn't "done" anything yet. Probably not a common use case, but if nothing else it's easier to describe/document the behavior.
2. By uploading the source artifacts right away, we'll be able to (re)use them for transforms.


gracecvking (approved):
👍

PR #54, opened 2023-02-08
Archives

mikeknep:
- Leverages the new support for tar.gz archive files as project artifacts to revamp how we manage relational artifacts:
```diff
- (N) source_{table}.csv
+ (1) source_tables.tar.gz

- (N) synth_{table}.csv
+ (1) synthetics_outputs.tar.gz
# Includes all synthetic output tables + all reports (individual and cross-table; html and json) + relational report (html)
```

- Renames evaluation outputs to better express their contents
```diff
- artifacts_{table}/  # subdir with report html, report json, logs
- expanded_evaluations_{table}.[html|json]
+ synthetics_cross_table_evaluation_{table}.[html|json]
+ synthetics_individual_evaluation_{table}.[html|json]
```

- Evaluation results (reports, scores) are now included in the restore process.

- Automatically run QualityReports/Evaluate for each table + create the relational report at the end of generate. This means users don't need to call `expand_evaluations` manually anymore (I've made it a private method, should be removed from examples/docs). Also the `create_report(multitable)` function is deprecated in favor of `multitable.create_relational_report()`.

- Small interface change to `MultiTable.init` that shouldn't affect users: instead of the restore workflow passing `project_unique_name`, we now pass a `restore_config` object (which includes the unique name + other data we need). The restore method constructs this config for you via the backup info, so you shouldn't ever need to make one and pass this argument yourself. (To consider: this does allow us to "un-deprecate" the `project_name` param; let's decide if we want to do that or keep the current/newer `project_display_name`.)

### DIFF
--- a/notebooks/relational.ipynb
+++ b/notebooks/relational.ipynb
@@ -252,7 +252,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "When training is complete, you'll find a number of artifacts in your working directory, including the CSVs on which models were trained (`train_{table}.csv`) and the standard Gretel model artifacts, including HTML and JSON reports and logs (`artifacts_{table}/`).\n",
+    "When training is complete, you'll find a number of artifacts in your working directory, including the CSVs on which models were trained (`synthetics_train_{table}.csv`) and evaluation reports (`synthetics_[type]_evaluation_{table}.[html|json]`).\n",
     "\n",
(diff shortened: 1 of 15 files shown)


gracecvking (approved):
🚢 it

gracecvking:
Just some quick documentation on what I saw, for my own sake. 

Independent synthesis. 4 tables. 
Console project artifacts view:
console

Downloaded `synthetics_outputs.tar.gz` 
downloaded synthetics_outputs tar gz

local working directory
local working dir

PR #59, opened 2023-02-14
Restore fixes

mikeknep:
- Fix missing keyword arg to `get_project`
- Prevent nested files in archives—add all files at archive root level
- Create working dir during restore if it doesn't exist (e.g. loading backup file from some other location)


gracecvking (approved):
LGTM

PR #60, opened 2023-02-16
Preprocessing updates

mikeknep:
Updates to how AncestralStrategy prepares training data:
- translates all keys (numeric or alphanumeric) to contiguous integers
- drops highly-NaN ancestral fields from training data (this is in addition to dropping highly-unique categorical ancestral fields, which we were already doing)
- adds new artificial records with min/max PK/FK values (instead of modifying existing records)


gracecvking (commented):
Let's have @amysteier do the approving review on this, since she knows the algorithm better than I do

amysteier (approved):
Looks good to me

PR #66, opened 2023-03-09
Bump minimum required python version to 3.8

mikeknep:
This unblocks us from updating a few dependencies (networkx, pandas, scikit-learn), but for now we'll stick to just the python min version update and can take on those dependency updates in separate PRs.


gracecvking (approved):
Let's just confirm with @johntmyers that this wont break anything

mikeknep:
Closing—we're going to go to 3.9 instead! 🚀

PR #67, opened 2023-03-09
Completely drop deprecated arguments and function

mikeknep:


gracecvking (approved):
👍 Nice
I've confirmed all current customers are using the most updated parameters, so we shouldnt have any incidents of notebooks erroring

PR #71, opened 2023-03-13
Relational eviction

mikeknep:
Adds project artifact eviction logic to Relational Trainer to prevent running into project artifact limits during runtime. We delete any project artifacts associated directly with a singular job (e.g. a model training dataset, or a seed file for conditional generation) when the job completes (or fails/is marked lost), so when "nothing is going on" we should expect to see some/all of the following artifacts (in order of when they'd be added, typically):
```
_gretel_backup.json
_gretel_debug_summary.json
source_tables.tar.gz
transforms_outputs.tar.gz
synthetics_training.tar.gz
synthetics_outputs.tar.gz
```
When we queue up jobs, once we reach the artifact limit we hold off on deferring additional jobs until there is more room in the project. This runs on the same `refresh_interval` check as other status updates. For example, if we have 8 jobs deferred, and on the next status check 2 jobs complete, we'll submit 2 of the 8 jobs and hold onto 6 until more space is available.

We do not upload a `transforms_training.tar.gz` archive since that is essentially identical to / a subset of `source_tables.tar.gz`.

If a notebook is interrupted during an Ancestral generation run and then restored (whether in progress or complete), it is possible/likely that some tables' seed files will have been already deleted, but they will not have been added to the archive because that happens at the very end of `generate`. The seed files should still exist locally, so they may still make it in to the archive after `mt = MultiTable.restore; mt.generate(resume=True)`, unless the restored instance is running in some other location where the original working directory is not present. We consider this acceptable loss, since the seed files are not particularly useful beyond potentially for debugging purposes.

### DIFF
--- a/src/gretel_trainer/relational/artifacts.py
+++ b/src/gretel_trainer/relational/artifacts.py
@@ -13,6 +13,7 @@ from gretel_client.projects import Project
 class ArtifactCollection:
     gretel_debug_summary: Optional[str] = None
     source_archive: Optional[str] = None
+    synthetics_training_archive: Optional[str] = None
     synthetics_outputs_archive: Optional[str] = None
     transforms_outputs_archive: Optional[str] = None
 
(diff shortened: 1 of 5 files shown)


gracecvking (approved):
LGTM

PR #76, opened 2023-03-16
Bump min required python to 3.9

mikeknep:


gracecvking (approved):
Hopefully colab doesnt move to 3.10 before we merge 😅

PR #77, opened 2023-03-16
Add passthrough policy to transform configs

mikeknep:
Relational Transforms requires user-supplied transform configs*. If such a config includes policies that match a primary or foreign key column name, those policies will be applied just like any other... but we don't want to do that, because Relational Transforms performs its own post-processing on key columns, and if any have been transformed by our models, we will have already corrupted the relationships between tables (and post-processing can't fix it).

This change modifies the transform configs supplied by the user, adding an initial passthrough policy that applies to all key columns on the table.

I tested this before and after with the [gdpr transforms config](https://raw.githubusercontent.com/gretelai/gdpr-helpers/main/src/config/transforms_config.yaml) and can confirm that on `main` we end up with no matching records between two tables in our `telecom` database (client and location), but when running this branch we preserve all PK/FK relationships between the two (an INNER join between the two tables does not drop any records).

Note: had to include some ugliness to support the two valid model keys (`transform` and `transforms`)—pretty sure this is unavoidable since we modify this config locally before submitting to the API, and both appear in various public templates/blueprints.

---

> \* Something to consider medium/longer-term: we could potentially default to using the gdpr transforms config linked above, instead of requiring users provide a transforms config. This would require changing the interface slightly, as currently the dict of `[table_name, xform_config]` passed to `train_transforms_models` currently dictates which tables we train transforms models.

### DIFF
--- a/src/gretel_trainer/relational/core.py
+++ b/src/gretel_trainer/relational/core.py
@@ -3,7 +3,8 @@ from __future__ import annotations
 import json
 import logging
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Optional
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Union
 
 import networkx
 import pandas as pd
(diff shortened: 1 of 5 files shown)


pimlock (approved):
Looks good!

PR #78, opened 2023-03-17
[PROD-386] Update relational report

tylersbray:
Setting this to draft:
* I need to test the `__repr__` I added for `TableEvaluation`
* See all the FIXME's floating around in `test_multi_table_restore.py`, I mocked with a sledge hammer but should have used a ball peen hammer.  Mike, I am kind of punting this one to you in the interest of not having it take forever, if you have pointers to fix these tests in a good way let me know.
* Now that I'm looking at this late on Friday, I see other stuff I could cut out for cleanliness (mostly css stuff).

I'll add some explanatory comments and questions throughout.

TableEvaluation takes only full json reports on instantiation, then it cherry picks values from the json (and NOT from the summary, there are no grades in there, only scores).  There are some changes around this sprinkled throughout non-reporting sections of the code.

`figures.py` is the relevant piece cut and pasted from monogretel.  I see that there are a couple lines I need to delete, I will do that before making this non-draft.

Grace's report.css had modified css for the tables.  I just nuked and paved it with monogretel's report.css.  I will also do a pass over that to delete anything that is obviously not needed for space, clarity and not leaking out too much.  Ditto the other css files, I just copied them over blindly to get things working but can probably cut out a lot.

Grace ok'd the html sample I gave her.  I'll produce a new one just in case I messed with anything since then, plus I added a second test to make a second report so I'll make that, too.  The scores are all mocked so might not look too different, but can't hurt.

Yes, I still owe you the javascript.


Makefile
```diff
@@ -9,6 +9,8 @@ type:
 .PHONY: multi
 multi:
 	-python -m pyright src/gretel_trainer/relational tests/relational/
+	# vvv Do we need this...?
+	# pip install -e .
```
tylersbray, line 13: I ran this manually a few time to get this make step to work, does it need to be enshrined forever or will the automated build be okay without it?

requirements-dev.txt
```diff
@@ -1,3 +1,6 @@
 -r requirements.txt
+black
```
tylersbray, line 2: New requirements for linting and for testing html.

requirements.txt
```diff
@@ -1,9 +1,11 @@
+backports.cached-property==1.0.0.post2
```
tylersbray, line 1: I used `@cached_property` just because some of it felt kind of silly (like stuff in templates where you do `if this then render this`).  Probably not a huge deal, if we don't want this dep I can just use plain properties everywhere.  I think cached_property is part of functools for python > 3.8?

src/gretel_trainer/relational/report/figures.py
```diff
@@ -0,0 +1,149 @@
+import math
+from typing import List, Optional
+
+import plotly.graph_objects as go
+
+_GRETEL_PALETTE = ["#A051FA", "#18E7AA"]
+_GRAPH_OPACITY = 0.75
```
tylersbray, line 7: Will delete these, not needed.

src/gretel_trainer/relational/report/report_template.html
```diff
@@ -3,91 +3,148 @@
...
+            <th>Primary Key</th>
+            <th>Foreign Keys</th>
+        </tr>
+        {% for table_data in presenter.report_table_data %}
+        <tr>
+          <td><b>{{ table_data.table }}</b></td>
+          <td><span class="label">{{ table_data.pk }}</span></td>
+          <td>
+            {% for fk in table_data.fks %}
+            <span class="label">{{ fk.column_name }}</span>&nbsp;
```
tylersbray, line 77: So with these label classes, js will add/remove label color classes on hover.  It will be great.

src/gretel_trainer/relational/core.py
```diff
@@ -19,9 +24,7 @@ class MultiTableException(Exception):
 @dataclass
 class TableEvaluation:
     cross_table_report_json: Optional[Dict] = field(default=None, repr=False)
-    cross_table_sqs: Optional[int] = None
```
gracecvking, line 22: @mikeknep What are your thoughts on this? I imagine we'd want to keep the sqs properties in, since they allow the scores to be viewed via `evaluations`.

src/gretel_trainer/relational/report/report.py
```diff
@@ -39,120 +53,94 @@ def generated_at(self) -> str:
...
+            if eval.individual_sqs is not None:
+                _total_score += eval.individual_sqs
+                _num_scores += 1
+            if eval.cross_table_sqs is not None:
+                _total_score += eval.cross_table_sqs
+                _num_scores += 1
+
+        # Take the average.
+        if _num_scores > 0:
+            score = int(_total_score / _num_scores)
```
gracecvking, line 71: Since int rounds, can we do the ceiling here instead?

src/gretel_trainer/relational/core.py
gracecvking, line 22: @mikeknep  What are your thoughts on this? I imagine we'd want to keep the sqs properties in, since they allow the scores to be viewed via evaluations.

src/gretel_trainer/relational/strategies/ancestral.py
```diff
@@ -283,7 +283,6 @@ def update_evaluation_from_model(
         common.download_artifacts(model, out_filepath, table_name)
 
         evaluation = evaluations[table_name]
-        evaluation.cross_table_sqs = common.get_sqs_score(model)
```
gracecvking, line 286: @mikeknep Wondering your thoughts on same as above

src/gretel_trainer/relational/strategies/independent.py
```diff
@@ -156,7 +156,6 @@ def update_evaluation_from_model(
         common.download_artifacts(model, out_filepath, table_name)
 
         evaluation = evaluations[table_name]
-        evaluation.individual_sqs = common.get_sqs_score(model)
```
gracecvking, line 159: @mikeknep also here

gracecvking (changes requested):
My only concern is the removal of the sqs properties in TableEvaluations makes running `multitable.evaluations` impossible (throws error). @mikeknep thoughts?

Makefile
mikeknep: I don't think it should be part of these make targets, but at some point we should consider some automation around the "welcome to Trainer" experience for devs/contributors.

The [CI build action](https://github.com/gretelai/trainer/blob/main/.github/workflows/build.yml#L17-L26) provides a good example of what needs to happen. (Maybe devs would want some virtualenv step.)

(Also just FYI as you can see in the build action linked above, in CI we run `make type` and `make test`, which cover the entire repo; these `multi[lint]` targets are just to speed up the local dev feedback loop while working exclusively on relational stuff.)

requirements.txt
mikeknep: I have a separate PR up to bump the minimum required Python version to 3.9, at which point you're right we'll be able to just pull this from `functools` and not need the backport. Maybe for this PR we remove this dep, switch to plain properties, and once 3.9 lands we update.

src/gretel_trainer/relational/core.py
mikeknep: The custom `__repr__` method that Tyler defined on this class (below) will ensure that we still see useful info in a notebook/repl when returning/printing `multitable.evaluations`.

src/gretel_trainer/relational/core.py
```diff
@@ -31,6 +34,62 @@ def is_complete(self) -> bool:
...
+
+    @property
+    def individual_ppl_grade(self) -> Optional[str]:
+        return self._grade_from_json(self.individual_report_json, _PPL)
+    
+    def __repr__(self) -> str:
+        d = asdict(self)
+        del d[cross_table_report_json]
+        del d[individual_report_json]
+        return json.dumps(d)
```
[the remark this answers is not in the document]
mikeknep: Also, just FYI the two `del` lines you have would need `self.` in front of the report json fields; but I think they'll come out anyways. Just saw it in the CI output.

tests/relational/test_multi_table_restore.py
```diff
@@ -215,6 +260,17 @@ def create_standin_project_artifacts(
...
+                html_filename = f"synthetics_{kind}_evaluation_{table}.html"
+                html_path = setup_path / html_filename
+                with open(html_path, "w") as f:
+                    f.write("<html></html>")
+                tar.add(html_path, arcname=f"{html_filename}")
+                json_filename = f"synthetics_{kind}_evaluation_{table}.json"
+                json_path = setup_path / json_filename
+                with open(json_path, "w") as f:
+                    json.dump(_report_json_dict, f)
+                tar.add(json_path, arcname=f"{json_filename}")
```
mikeknep, line 273: I probably didn't explain this well enough / it isn't clear enough from code (😞). For synthetics, there are two archive files we create, at the following times, with the following contents:

**synthetics_training.tar.gz**
- Created upon completion of synthetics models training (mt.train)
- Contains the CSVs used as data sources for each table's model


**synthetics_outputs.tar.gz**
- Created upon completion of synthetics record handlers + post-processing (mt.generate)
- Contains, at the root level of the archive, N subdirectories, one for each "run" (each call to mt.generate) (since you may want to train once and generate N times). Each run subdirectory contains:
  - relational report
  - all component report files (4 per table: individual and cross-table, html and json)
  - synthetic output data as CSVs
  - conditional seed CSVs (if any)

So we shouldn't add these files to the training archive in this test. (For more context: at training-complete-time, we'd only have one set of those files available—the artifacts downloaded from the model. Which "kind" those artifacts are (individual or cross-table) depends on the strategy used.)

tests/relational/test_multi_table_restore.py
```diff
@@ -386,8 +436,9 @@ def test_restore_training_complete(project, pets):
         mt = MultiTable.restore(backup_file)
 
         # Backup + Debug summary + Source archive + (2) Source CSVs
-        # + Training archive + (2) Train CSVs + (4) Reports
-        assert len(os.listdir(working_dir)) == 12
+        # + Training archive + (2) Train CSVs + (4) Train reports + (4) Reports
+        # FIXME verify summary above is correct
+        assert len(os.listdir(working_dir)) == 16
```
mikeknep, line 441: See above, but the original count here is correct, we'd expect 12:
- the extracted training archive would only have 2 CSVs (zero reports)
- the 4 report files are (re-)downloaded directly from the model during restore

tests/relational/test_multi_table_restore.py
```diff
@@ -443,19 +498,23 @@ def test_restore_training_one_failed(project, pets):
 
         # Backup + Debug summary + Source archive + (2) Source CSVs
         # Training archive + (2) Train CSVs + (2) Reports
-        assert len(os.listdir(working_dir)) == 10
+        # FIXME this was 10, not sure how we got 16 if something "failed" because we might want like 14 then? 12?
+        assert len(os.listdir(working_dir)) == 16
```
mikeknep, line 502: 1. backup
2. debug summary
3. source archive
4. humans source CSV (extracted from source archive)
5. pets source CSV (extracted from source archive)
6. training archive
7. humans train CSV (extracted from training archive)
8. pets train CSV (extracted from training archive)
9. pets report HTML (downloaded from pets model)
10. pets report JSON (downloaded from pets model)

in this test, the humans model failed to train, so we can't/don't download anything from it.

mikeknep (commented):
Love what you've done here!!

I'll dig deeper into the restore test—don't worry about the comments there, they're just FYI

mikeknep:
I fixed the restore test setup, and while I was there I knocked a few of these other things out of the way 😀  I made a separate branch and PR for you to check out, instead of just stomping on top of your branch directly. https://github.com/gretelai/trainer/pull/79

requirements.txt
mikeknep: Update: I merged in the 3.9 requirement, so if you rebase/merge latest `main` you'll be able to import this from functools 🎉

Makefile
tylersbray: Removed via Mike's branch.

requirements.txt
tylersbray: Rebased and updated import, thanks.

src/gretel_trainer/relational/report/figures.py
tylersbray: Deleted, also trimmed out a bunch of unneeded css.

src/gretel_trainer/relational/report/report.py
tylersbray: Using ceil now.

src/gretel_trainer/relational/report/report_template.html
```diff
@@ -3,91 +3,148 @@
...
+          <h1>Gretel Relational<br />Synthetic Report</h1>
+          <span>
+            <b>Generated</b> {{presenter.generated_at}}
+            <br/>
+            <b>Generation ID</b> {{ presenter.run_identifier }}
+          </span>
+        </div>
+
+        <div class="score-container mb-1">
+            <span class="label label__{{ presenter.composite_sqs_score_and_grade[1].lower().replace(' ', '-') }} mb-0.5">
```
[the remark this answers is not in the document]
tylersbray: Added a little something, template is cleaner now.

src/gretel_trainer/relational/report/report_template.html
```diff
@@ -3,91 +3,148 @@
...
+              <span class="label label__{{ presenter.sqs_score_to_grade(evaluation.individual_sqs).lower().replace(' ', '-') }}">
+                {{ presenter.sqs_score_to_grade(evaluation.individual_sqs) }}
+              </span>
+              <span class="sqs-table-link">
+                <a href="synthetics_individual_evaluation_{{ table }}.html">Report</a>
+              </span>
+            </td>
+            <td>
+              <span class="sqs-table-score">{{ evaluation.cross_table_sqs }}</span>
+              <span class="label label__{{ presenter.sqs_score_to_grade(evaluation.cross_table_sqs).lower().replace(' ', '-') }}">
```
mikeknep, line 127: Could these two (top and bottom of highlighted lines) use the label-maker method as well? (I guess the helper method would have to get reworked a bit since it's calling the same formatting stuff on different properties.)

mikeknep (approved):
Just one more nit, since you so kindly obliged my earlier non-blocking suggestion. Thanks for this!!

tylersbray:
Subsumed by https://github.com/gretelai/trainer/pull/81

PR #80, opened 2023-03-20
Relax dependency requirements

mikeknep:
Previously we pinned explicitly down to exact patch numbers; now we define compatible release clauses to ensure we don't unknowingly bump a **major** version, but allow variance at the minor and patch release levels. This should make it easier for clients to install and run Trainer alongside whatever other libraries exist in their environments (e.g. particularly in Colab, which pre-installs a bunch of packages for convenience).

I did pull in a major version update to `networkx` here; our use of that library is pretty confined to very core stuff that [hasn't changed significantly](https://networkx.org/documentation/latest/release/migration_guide_from_2.x_to_3.0.html), and our unit tests for the class using it (`RelationalData`) all still pass.

The only type hint we were importing from `typing_extensions` was Protocol, which was added to `typing` core in Python 3.8. Now that we require >= 3.9, we can drop this dep and import directly from `typing`.


requirements.txt
```diff
@@ -1,12 +1,11 @@
...
-scikit-learn==1.0.2
-smart-open[s3]==5.2.1
-sqlalchemy==1.4.42
-typing-extensions==4.3.0
+jinja2~=3.1
+networkx~=3.0
+pandas~=1.3
+pydantic~=1.9
+requests~=2.25
+scikit-learn~=1.0
```
gracecvking, line 9: Relaxing the scikit requirement fixed the colab but we've been hitting with relational notebooks 😁 💯 .

gracecvking, line 9: Relaxing the scikit requirement fixed the colab but we've been hitting with relational notebooks 😁 💯 .

tylersbray (approved):
lgtm

PR #83, opened 2023-03-21
Parallel evaluate and task refactor

mikeknep:
### User-facing feature

Previously, when synthetic model record handlers all competed (within `multitable.generate`), we ran synchronous Evaluate jobs using the `QualityReport` helper class to get expanded evaluation metrics.

This PR parallelizes that work by ditching the `QualityReport` helper and instead creating standard Evaluate jobs "manually" and running them in the same "Gretel job execution loop" pattern as other workflow tasks (i.e. training and running transforms and synthetics models).

### Tech debt refactor

The bulk of the diff here is the result of introducing an abstraction for that "Gretel job execution loop" described above. The generic infrastructure is now defined in `src/gretel_trainer/relational/task_runner.py`—start jobs when you can, check if there's more to do, update statuses and handle various states, etc. Then, `src/gretel_trainer/relational/tasks/*.py` defines the details for each concrete task (in "chronological" order: TransformsTrain, TransformsRun, SyntheticsTrain, SyntheticsRun, SyntheticsEvaluate).


src/gretel_trainer/relational/model_config.py
```diff
@@ -14,6 +14,12 @@ def _model_name(workflow: str, table: str) -> str:
     return f"{workflow}-{ok_table_name}"
 
 
+def make_evaluate_config(table: str) -> Dict[str, Any]:
```
gracecvking, line 17: This is nice because _down the line_ if we decide to include more evaluate tasks (I've been chatting very preliminarily with Nicole about this potential), we could edit that here.

src/gretel_trainer/relational/tasks/common.py
```diff
@@ -0,0 +1,29 @@
...
+from gretel_trainer.relational.strategies.independent import IndependentStrategy
+
+
+class _MultiTable(Protocol):
+    @property
+    def _refresh_interval(self) -> int:
+        ...
+
+    @property
+    def _project(self) -> Project:
```
gracecvking, line 17: Oooo does this mean we could add project name/link to relational report?
(Don't worry, I'm not trying to do this now)

gracecvking (approved):

src/gretel_trainer/relational/tasks/common.py
mikeknep: Yep that's definitely available if we want it on the report.

PR #84, opened 2023-03-21
Use legacy resolver arg for pip install in benchmark notebook

misberner:
At least on my colab instance (and I guess on most environments), the `gretel-trainer` dep won't install without telling `pip` to use the legacy resolver.


misberner:
Confirmed that this is no longer needed with the current `main` version

PR #85, opened 2023-03-21
Docs

mikeknep:
We added `jinja2` as a dependency to Trainer's main `requirements.txt` file a while back, causing a conflict with the version required here for readthedocs. This PR drops the reference to that full requirements file, as well as requests and gretel-client, which I think are pretty obviously not needed for building markdown docs.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,9 +1,9 @@
 -r ../requirements.txt
-m2r
-sphinx==3.0.3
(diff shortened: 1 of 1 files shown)


gracecvking:
I just wrote myself a note to ask you @mikeknep about whether trainer has readthedocs. As usual, you're one step ahead of me!

tylersbray (approved):
Thanks!

PR #86, opened 2023-04-11
Hybrid trainer

mikeknep:
#### Hybrid support
- Use the generic `submit` method on jobs (which relies on the session config, see more below)
- Skip artifact eviction when in hybrid mode (we can't delete hybrid artifacts, but we don't need to anyways because we don't have an artifact limit to worry about)

#### Lift configure_session out of Trainer constructor
Cloud vs. Hybrid is driven by the client config. Lifting this function out of the library code separates concerns and avoids cluttering the Trainer interface with configuration details. I updated all the (non-Benchmark/Relational) notebooks and examples to explicitly call `configure_session` first separately. (Relational coming in a separate PR.)

### DIFF
--- a/notebooks/conditional-generation.py
+++ b/notebooks/conditional-generation.py
@@ -1,4 +1,5 @@
 import pandas as pd
+from gretel_client import configure_session
 
 from gretel_trainer import Trainer
 from gretel_trainer.models import GretelLSTM, GretelACTGAN
@@ -19,6 +20,9 @@ seed_df = pd.DataFrame(data=[
     ["asian", "chinese", "F"]
 ], columns=["RACE", "ETHNICITY", "GENDER"])
 
+
+# Configure Gretel credentials
+configure_session(api_key="prompt", cache="yes", validate=True)
 
 # Train a model and conditionally generate data
 seed_fields = seed_df.columns.values.tolist()
--- a/notebooks/custom-example.py
+++ b/notebooks/custom-example.py
@@ -1,6 +1,11 @@
+from gretel_client import configure_session
+
 from gretel_trainer import Trainer
 from gretel_trainer.models import GretelLSTM, GretelACTGAN
 
+
+# Configure Gretel credentials
+configure_session(api_key="prompt", cache="yes", validate=True)
 
 dataset = "https://gretel-public-website.s3-us-west-2.amazonaws.com/datasets/USAdultIncome5k.csv"
 
--- a/notebooks/simple-conditional-generation.ipynb
+++ b/notebooks/simple-conditional-generation.ipynb
@@ -32,16 +32,28 @@
     },
     {
       "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "# Configure Gretel credentials\n",
+        "from gretel_client import configure_session\n",
+        "\n",
+        "configure_session(api_key=\"prompt\", cache=\"yes\", validate=True)"
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "Gh10cM4SiHPM"
+      },
+      "outputs": [],
       "source": [
         "# Train model\n",
         "model = trainer.Trainer()\n",
         "model.train(DATASET_PATH, seed_fields=SEED_FIELDS)"
-      ],
-      "metadata": {
-        "id": "Gh10cM4SiHPM"
-      },
-      "execution_count": null,
-      "outputs": []
+      ]
     },
     {
       "cell_type": "code",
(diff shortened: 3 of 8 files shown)


notebooks/conditional-generation.py
```diff
@@ -20,6 +21,15 @@
...
 
 
+# Configure Gretel credentials
+configure_session(
+    api_key="prompt",
+    cache="yes",
+    validate=True,
+    # default_runner="cloud",
+    # artifact_endpoint="cloud",
+)
```
mikeknep, line 31: I don't have a strong feeling on how we format this specifically, but I do think it'd be nice to be consistent* across all our notebooks both in this repo and in the UCC (relational). As currently written, this presents additional optional arguments with their default values commented out. Would it be more useful to show what hybrid values would look like? Something like:
```python
configure_session(
    api_key="prompt",
    cache="yes",
    validate=True,
    # uncomment and complete below for hybrid contexts
    # default_runner="hybrid",
    # artifact_endpoint="s3://my-bucket", # replace with your bucket name
)
```


*Note: there may be cases where we're explicitly demoing how to configure a hybrid session, in which case obviously the presentation of `configure_session` will be different. In general, though, I think it'd be good to have a standard style for generic contexts.

gracecvking: The majority of our notebooks today use one of the two following formats:
`configure_session(api_key="prompt", cache="yes", validate=True)`
or
`configure_session(endpoint="https://api.gretel.cloud", api_key="prompt", cache="yes")`

I think we should have the common format be the former - not show commented out defaults. And add explicit instructions in the docs about how to configure for hybrid (and maybe a text cell pointing those interested in hybrid to the docs)

@sami-t I'm curious if you have strong feelings on this

gracecvking (changes requested):
The majority of our notebooks today use one of the two following formats:
configure_session(api_key="prompt", cache="yes", validate=True)
or
configure_session(endpoint="https://api.gretel.cloud", api_key="prompt", cache="yes")

I think we should have the common format be the former - not show commented out defaults. And add explicit instructions in the docs about how to configure for hybrid (and maybe a text cell pointing those interested in hybrid to the docs)

@sami-t I'm curious if you have strong feelings on this

sami-t:
> The majority of our notebooks today use one of the two following formats: configure_session(api_key="prompt", cache="yes", validate=True) or configure_session(endpoint="https://api.gretel.cloud", api_key="prompt", cache="yes")
> 
> I think we should have the common format be the former - not show commented out defaults. And add explicit instructions in the docs about how to configure for hybrid (and maybe a text cell pointing those interested in hybrid to the docs)
> 
> @sami-t I'm curious if you have strong feelings on this

No strong feelings about the specific format, as long as (1) we're consistent across notebooks, and (2) we don't require users to hunt down how to do hybrid configuration. That could mean either adding a text cell (as you suggested @gracecvking) in the notebook pointing folks to specific hybrid instructions from which they can quickly copy/paste a hybrid version of `configure_session` (without having to read a long doc), or adding commented out `default_runner` and `artifact_endpoint` params as you suggested @mikeknep.

mikeknep:
I'm leaning towards a single cell purely for copy/paste consistency, and also our notebooks vary in terms of amount of prose. I feel like this might strike the right balance of making all three of "keep consistent across notebooks" + "adjust to hybrid" + "learn more about hybrid" pretty easy.
```python
configure_session(
    api_key="prompt",
    cache="yes",
    validate=True,
    # uncomment and adjust below for hybrid; more info at https://docs.gretel.ai/...
    # default_runner="hybrid",
    # artifact_endpoint="s3://my-bucket", # replace with your artifact endpoint
)
```
Of course, this requires actually writing some docs about hybrid, which we don't have yet... Maybe for now we should just stick with the three main args, no comments or links or anything yet, and we can sweep through with another update when we have public hybrid docs to reference.

(Side note: this has me wishing notebooks supported something like front-end components where we could define this in one central place and reference it in multiple notebooks. Alas.)

mikeknep:
Update: as suggested in my last comment, I set all the examples to just call `configure_session(api_key="prompt", cache="yes", validate=True)` for now, and we can decide if-and-how we want to change that for hybrid once we have public docs available.

notebooks/conditional-generation.py
drew: +1 for keeping the original boilerplate. I'd optimize for keeping blueprint boilerplate consistent, concise and readable.

Getting a hybrid environment setup will require quite a bit of extra commitment and setup regardless, so I think asking these customers to update their boilerplate isn't too much of an ask. We'll also need to provide additional guidance such as configuring their local environment with access to CP object stores, so even adding the commented out config options isn't sufficient to setup a hybrid environment.

gracecvking (approved):

PR #87, opened 2023-04-12
Hybrid relational

mikeknep:
### Similarities to [Trainer PR](https://github.com/gretelai/trainer/pull/86)
These changes will also require gretel-client 0.16.0, which I specified in the requirements file in the PR for trainer. I'll rebase this PR atop that one once it lands.

As with Trainer, I've lifted `configure_session` out of the library code. We should add a `configure_session` cell to the UCC notebooks.

### Relational artifacts
We do not upload relational-specific artifacts in hybrid mode. This includes the debug summary, backup file, and archive files (source, xforms outputs, synthetics training, synthetics outputs). The reason for this is that all of these relational artifacts are intended to be "singleton" objects that get overwritten when/as state changes. Examples:
- backup file: we take backup snapshots quite often throughout all processes
- xforms/synthetics outputs archives: each time you run record handlers (e.g. synthetics `generate`) we add that run's data to a subdirectory in the archive; the full history of runs is stored in the tar file
- source archive: changes the least often, but it will change if you run transforms with `in_place=True` or if you retrain synthetics tables

In hybrid contexts, we cannot list or delete project artifacts, so there is no way for us to _overwrite_ these files in the user's bucket. We'd end up polluting their bucket with a ton of nearly-duplicate-named files (e.g. `gretel_{uuid}_synthetics_outputs.tar.gz` and an inordinate amount of `gretel_{uuid}__gretel_backup.json` files) with no obvious way to know which is the latest.

Note: this limits the restore functionality a bit. In cloud context, we have artifact IDs for these and can download them from the project, so as long as you have a backup JSON file on hand you can restore "from anywhere", even a separate machine. In hybrid context, we are reliant on the files in the local working directory.

### Refactoring diffs
A good chunk of the diff here is due to grouping all the functions from `sdk_extras.py` into a class. (This prevents us from having to pass around `hybrid: bool` flags all over the place.)

I also introduced a pair of fixtures to the restore unit tests to reduce duplication (and save a level of indentation 😃), so that file looks noisier than it really is—none of the tests changed functionally (and one was added).

### DIFF
--- a/src/gretel_trainer/relational/artifacts.py
+++ b/src/gretel_trainer/relational/artifacts.py
@@ -11,6 +11,7 @@ from gretel_client.projects import Project
 
 @dataclass
 class ArtifactCollection:
+    hybrid: bool
     gretel_debug_summary: Optional[str] = None
     source_archive: Optional[str] = None
     synthetics_training_archive: Optional[str] = None
@@ -37,7 +38,10 @@ class ArtifactCollection:
         existing = self.transforms_outputs_archive
         self.transforms_outputs_archive = self._upload_file(project, path, existing)
 
-    def _upload_file(self, project: Project, path: str, existing: Optional[str]) -> str:
+    def _upload_file(self, project: Project, path: str, existing: Optional[str]) -> Optional[str]:
+        if self.hybrid:
+            return None
+
         latest = project.upload_artifact(path)
         if existing is not None:
             project.delete_artifact(existing)
--- a/src/gretel_trainer/relational/artifacts.py
+++ b/src/gretel_trainer/relational/artifacts.py
@@ -41,6 +41,9 @@ class ArtifactCollection:
     def _upload_file(
         self, project: Project, path: str, existing: Optional[str]
     ) -> Optional[str]:
+        # We do not upload any of these artifacts in hybrid contexts because they are intended to be
+        # "singleton" objects, but we cannot list or delete items in users' artifact endpoints, so
+        # we would end up polluting their endpoints with many nearly-duplicative copies of these files.
         if self.hybrid:
             return None
 
--- a/src/gretel_trainer/relational/artifacts.py
+++ b/src/gretel_trainer/relational/artifacts.py
@@ -38,7 +38,9 @@ class ArtifactCollection:
         existing = self.transforms_outputs_archive
         self.transforms_outputs_archive = self._upload_file(project, path, existing)
 
-    def _upload_file(self, project: Project, path: str, existing: Optional[str]) -> Optional[str]:
+    def _upload_file(
+        self, project: Project, path: str, existing: Optional[str]
+    ) -> Optional[str]:
         if self.hybrid:
             return None
 
--- a/src/gretel_trainer/relational/sdk_extras.py
+++ b/src/gretel_trainer/relational/sdk_extras.py
@@ -1,4 +1,5 @@
 import logging
+import shutil
 from contextlib import suppress
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
@@ -53,8 +54,10 @@ class ExtendedGretelSDK:
     ) -> None:
         download_link = gretel_object.get_artifact_link(artifact_name)
         try:
-            with open(out_path, "wb+") as out:
-                out.write(smart_open.open(download_link, "rb").read())
+            with smart_open.open(download_link) as src, smart_open.open(
+                out_path, "w"
+            ) as dest:
+                shutil.copyfileobj(src, dest)
         except:
             logger.warning(f"Failed to download `{artifact_name}`")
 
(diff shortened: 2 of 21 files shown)


gracecvking (approved):
LGTM. I appreciate the detailed explanation in your description!

src/gretel_trainer/relational/sdk_extras.py
```diff
@@ -18,90 +18,92 @@
...
+    def download_file_artifact(
+        self,
+        gretel_object: Union[Project, Model],
+        artifact_name: str,
+        out_path: Union[str, Path],
+    ) -> None:
+        download_link = gretel_object.get_artifact_link(artifact_name)
+        try:
+            with open(out_path, "wb+") as out:
+                out.write(smart_open.open(download_link, "rb").read())
```
johntmyers, line 57: May want to consider using `shutil.copy()` or something to avoid memory bloat on large artifacts, see: https://github.com/Gretellabs/monogretel/pull/2036

src/gretel_trainer/relational/strategies/common.py
```diff
@@ -9,12 +9,14 @@
...
-from gretel_trainer.relational.sdk_extras import download_file_artifact
+from gretel_trainer.relational.sdk_extras import ExtendedGretelSDK
 
 logger = logging.getLogger(__name__)
 
 
-def download_artifacts(model: Model, out_filepath: Path, table_name: str) -> None:
+def download_artifacts(
+    model: Model, out_filepath: Path, table_name: str, extended_sdk: ExtendedGretelSDK
+) -> None:
```
johntmyers, line 19: docstrings need updated, no return type

src/gretel_trainer/relational/strategies/common.py
```diff
@@ -9,12 +9,14 @@
...
 
 from gretel_trainer.relational.core import RelationalData
-from gretel_trainer.relational.sdk_extras import download_file_artifact
+from gretel_trainer.relational.sdk_extras import ExtendedGretelSDK
 
 logger = logging.getLogger(__name__)
 
 
-def download_artifacts(model: Model, out_filepath: Path, table_name: str) -> None:
+def download_artifacts(
```
johntmyers, line 17: What is `table_name` used for in this function? I didn't see any refs

src/gretel_trainer/relational/sdk_extras.py
```diff
@@ -18,90 +18,92 @@
...
+            project, number_of_artifacts
+        ):
+            self._log_start(table_name, action)
+            job.submit()
+        else:
+            self._log_waiting(table_name, action)
+
+    def _room_in_project(self, project: Project, count: int) -> bool:
+        if self._hybrid:
+            return True
```
johntmyers, line 100: mmmmm free storage 🤤

src/gretel_trainer/relational/artifacts.py
```diff
@@ -37,7 +38,12 @@ def upload_transforms_outputs_archive(self, project: Project, path: str) -> None
         existing = self.transforms_outputs_archive
         self.transforms_outputs_archive = self._upload_file(project, path, existing)
 
-    def _upload_file(self, project: Project, path: str, existing: Optional[str]) -> str:
+    def _upload_file(
+        self, project: Project, path: str, existing: Optional[str]
+    ) -> Optional[str]:
+        if self.hybrid:
+            return None
```
johntmyers, line 45: nit: this makes sense given the description in the PR, about polluting the hybrid storage buckets, but might be worth adding some details about that here as it's not obvious by just looking at this code why we skip the upload for hybrid

src/gretel_trainer/relational/strategies/common.py
mikeknep: Nice catch. I think it had been used previously, but a refactor made it obsolete and I missed the signature change. The callers of this function use `table_name` to compose `out_filepath`, and no longer need to pass the string down here. Fixing!

src/gretel_trainer/relational/artifacts.py
mikeknep: Good call 👍

src/gretel_trainer/relational/sdk_extras.py
mikeknep: I don't think we can use `shutil` because `download_link` is not a local path—it's either a cloudfront URL (cloud) or an S3/etc. URL (hybrid).

We may want to consider adding first class support for downloading individual Gretel artifacts to the client. Currently we have `download_artifacts` on Jobs, but of course that downloads _all_ artifacts and we only need one or two. There is no download method for project artifacts, only `get_artifact_link`.

If we did add new `download_artifact` (singular) methods to Projects and Jobs in the client, we could use `requests` with the Cloudfront URL in the cloud context to avoid reading the artifact, but I think in hybrid we'd still be stuck using `smart_open` to handle abstracting over the specific cloud provider object storage being used.

johntmyers: But `smart_open` gives you the file-like object that `shutil.copyfileobj` should be able to use.

mikeknep: 🤯

PR #88, opened 2023-04-13
Trainer types

mikeknep:
While working on hybrid support, I thought I'd tackle the pyright warnings so that we can type check the entire codebase (previously had been limited to Benchmark and Relational). Some notes:
- I punted on some of the Benchmark warnings (`# type:ignore`) because Benchmark v2 is on the horizon
- I didn't see any public method on `StrategyRunner` that didn't ultimately call the `needs_load` wrapper, so I decided to just refactor such that that is called during init instead of decorating basically every method
- Dropped `StrategyRunner.from_completed` as it is unused (perhaps it was supplanted by `Trainer.load`, which appears to do the same thing)
- I didn't see anything setting `PartitionConstraints max_row_partitions`. Not sure if this had been used and was deprecated/removed, or if it had never been used; in any case, I removed it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ test:
 
 .PHONY: type
 type:
-	python -m pyright src/gretel_trainer/benchmark src/gretel_trainer/relational tests
+	python -m pyright src tests
 
 .PHONY: multi
 multi:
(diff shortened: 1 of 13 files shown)


drew (approved):
Awesome!

pimlock (approved):
💯

PR #89, opened 2023-04-14
Initial actions PoC

johntmyers:


actions/data/transform_config.yml
mikeknep, line 1: Lately I've been using the GDPR transforms config as my go-to default. Could we use that config here? Admittedly it's a bit clunky at the moment (referencing it via a raw github URL) but I think the plan is to eventually bring it in to blueprints and give it a "friendly alias".

(OTOH, it may be nice to keep this around if for no other reason than to effectively demo providing a custom config. I don't feel strongly about it either way.)

actions/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Gretel API Key
+- The password for the MySQL db
+
+```
+GRETEL_CONFIG=/gretel/data/transform_config.yml
+GRETEL_API_KEY=grtuXXX
+SOURCE_DB=mysql://dynf4plkdpg6e14abhf5:PASSWORD@aws.connect.psdb.cloud:3306/relational-demo?ssl=true&charset=utf8mb4
+SINK_DB=sqlite:////gretel/data/transformed.db
+```
```
mikeknep, line 42: Should we add `.env` to the repo's gitignore just to be safe?

Also, is it worth checking in a `.env.example` file and instructing users here to `cp .env.example .env` and then adjust the newly-created file? I've always kind of liked that pattern.

actions/data/transform_config.yml
mikeknep: I guess this also serves as a functional demo of mounting in to `/gretel/data` so yeah probably keep it around.

src/gretel_trainer/relational/multi_table.py
```diff
@@ -5,29 +5,23 @@
...
 import shutil
 import tarfile
-import time
 from collections import defaultdict
-from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import suppress
-from dataclasses import dataclass, field, replace
+from dataclasses import replace
 from datetime import datetime
-from enum import Enum
```
mikeknep, line 14: Sigh, I need to fix my vim LSP so that I'd notice all these unused imports 😭

mikeknep (approved):
Very cool! 😎

actions/README.md
johntmyers: It's already in `.gitignore` all default Python ignore files from GitHub already contain that

johntmyers: On the second note, yeah, I have a `.env.template` I didn't commit that has all the possible envs we support and was going to add some comments for how they are used. 

Arguably, any action could use any subset of all env vars depending on what's being done. For example Relational wouldn't make use of `GRETEL_PROJECT`, but a vanilla trainer based action might for just training a model + synthetic data output

PR #101, opened 2023-04-27
Seed fix

mikeknep:
Fixes two bugs.

1. When using Ancestral strategy, we were previously basing the "highly NaN" and "highly unique categorical" calculations (to determine whether or not we include them in training and seeding) on the column data as it appeared in the joined ancestral, not the source parent table. A column could be highly unique in the source (e.g. team name in the NBA database `Teams` table), but as a result of the particular foreign key frequency on some child table it is possible to _not_ hit the uniqueness threshold when appearing as a joined ancestral column. This led to a bug where we would train on that data and attempt to use it as a seed column, but the synthetic child data used as a seed would include categorical values not seen during training, ultimately resulting in having to throw away records as invalid conditional seeds.
2. When label encoding keys on tables, we were previously iterating through the tables in an arbitrary order. This led to a bug (possibly limited to composite keys): a child table with a composite PK made up of two individual FK columns could be label-encoded "in isolation" before the parent table is label-encoded, and as a result we lost referential integrity between the two. We now ensure that tables are label-encoded in proper order, i.e. parents before children.

### Implementation note

The logic for determining if a column is highly NaN or highly unique categorical now exists in the `RelationalData` class itself, rather than being owned by the `AncestralStrategy`. One immediate benefit is we now perform the calculation just once for each table and cache the result, rather than recalculate it every time the table is joined onto a child table. Additionally, in the future there _might_ be ways for us to get this metadata from a more sophisticated connector... **if** that is possible, the result would need to live in `RelationalData` anyways, so making this change now sets us up better for that possibility.

### DIFF
--- a/src/gretel_trainer/relational/ancestry.py
+++ b/src/gretel_trainer/relational/ancestry.py
@@ -50,6 +50,7 @@ def get_table_data_with_ancestors(
     rel_data: RelationalData,
     table: str,
     tableset: Optional[Dict[str, pd.DataFrame]] = None,
+    ancestral_seeding: bool = False,
 ) -> pd.DataFrame:
     """
     Returns a data frame with all ancestral data joined to each record.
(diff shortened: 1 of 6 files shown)


gracecvking (approved):
👍

PR #102, opened 2023-04-28
Classify

mikeknep:
Adds support for classify!

- At the public interface level, just a single method: `classify`
- Takes one classify config and applies it to all tables
- By default, only creates and trains models. Classify training produces a "data preview" that is the classification results for the first 100 records from the data source. In many cases this is all a user needs to identify columns of interest/concern.
- Optionally pass `all_rows=True` to run record handler jobs after training completes; the data results contain all records.
- Results are put in a new archive file called `classify_outputs.tar.gz`, which is uploaded to the project
  - By default, this archive will only contain `classify_{table}.gz` files
  - If `all_rows=True`, the archive will contain ^^ plus `classify_all_rows_{table}.gz` files
- We don't load/expose the results into memory—all we do is download the data result files for the user and put them in the working directory. There are two main reasons for this:
  - There isn't an especially obvious/convenient way to display the results (which are in JSONL format)—we could parse into a list of dictionaries, but bleh; it'd also be gigantic output (Maarten prints the first 5 in his notebook)
  - We want to avoid unnecessarily loading a bunch of data into memory
- The Classify model IDs are kept in memory so if you call `classify()` a second time we don't retrain. This seems particularly useful in the case of a user looking at the subset results (from training only) and wanting to subsequently run with `all_rows=True`—in that case they'll jump straight to the record handler and not have to retrain.
  - This is also supported through the restore workflow; in other words, we include classify model IDs in the backup data. As usual, if a classify outputs archive file exists in the project we will download and unpack it (cloud only, not hybrid).

There's quite a bit of additional noise here because I refactored the `Task` abstraction and `run_task` function, which meant slightly tweaking all the current tasks (synthetics train, transforms run, etc.). These changes were primarily to support a single Task (classify) being able to run both models and record handlers. This sets us up to potentially consolidate current multi-method workflows into single-method workflows, for example:
```
train_transform_models + run_transforms => transform
train + generate => synthesize
```

### DIFF
--- a/notebooks/relational.ipynb
+++ b/notebooks/relational.ipynb
@@ -249,6 +249,57 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "#### Classify\n",
+    "\n",
+    "Run Gretel Classify on all tables to identify PII.\n",
+    "\n",
+    "By default, Relational Classify will provide results for the first 100 rows in each table. To process the entire table, set `all_rows=True`."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import yaml\n",
+    "\n",
+    "config_yaml = \"\"\"\n",
+    "schema_version: \"1.0\"\n",
+    "name: \"classify-default\"\n",
+    "models:\n",
+    "  - classify:\n",
+    "      data_source: \"_\"\n",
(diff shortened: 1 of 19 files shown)


gracecvking (approved):
Re: finalizing naming for results put in `classify_outputs.tar.gz`
Let's call `classify_subset_{table}.gz` --> `classify_{table} and keep `classify_all_rows_{table}` as is.

PR #104, opened 2023-05-03
Fix FKs in relational report

mikeknep:
I missed this in #100 when we turned all the keys into lists 😞  Fortunately we haven't released yet so this is only broken on `main` branch, not in the latest public version on PyPi.


tylersbray (approved):
Makes sense and looks good, thanks.

PR #107, opened 2023-05-11
Support for JSON

mikeknep:
### How it works

Whenever a table is added (via `add_table`), we check to see if any columns in the table contain structured data—this can be actual dictionaries/lists or a valid JSON string. If it does, we _invent_ 1-N tables that flatten the nested data by lifting dictionary keys to standalone columns and moving lists to new child tables. We train using all the flattened tables, and at the very end reassemble generated data to the original nested format.

### What's user-facing?

This is designed to be nearly completely transparent to users, but there are some public-facing things to note:
- In Console, you'll see models for invented tables
- In the Relational Report, we do not show any invented tables. The SQS score comes from running Evaluate on the root invented table only. Same for the `evaluations` dict property on `MultiTable`.
- The `list_all_tables` method now takes an optional "scope" parameter to return different subsets of the known tables. By default we **will** expose the invented table names and **not** the original table name since this method is used so frequently by `MultiTable` (which generally needs exactly that), but users can get their original table names by passing `scope="public"` | `scope=Scope.PUBLIC`.

All the manual modification methods (e.g. `set_primary_key`, `remove_foreign_key`, etc.) have been updated to work correctly when referencing a source-with-JSON table. The public interface is unchanged. (This accounts for the majority of the changes in the `relational.core` module.)


src/gretel_trainer/relational/json.py
```diff
@@ -0,0 +1,310 @@
...
+def get_parent_column_name_from_child_table_name(table_name: str) -> str:
+    return table_name.split(TABLE_SEPARATOR)[-1]
+
+
+def is_child_table(df: pd.DataFrame) -> bool:
+    id_columns = get_id_columns(df)
+    if len(id_columns) != 1:
+        return False
+    id_col = id_columns[0]
+    return all([col in df.columns for col in (id_col, ORDER_COLUMN, CONTENT_COLUMN)])
```
sami-t, line 118: Now that you're keeping track of metadata, is there a way to leverage that to make these functions less hacky? E.g. could we use `_is_invented` instead of `is_child_table` without a major refactoring?

sami-t (approved):

src/gretel_trainer/relational/json.py
mikeknep: Good idea, thanks. I was able to remove a few of these entirely, and replace most (but not all) uses of the others. We do need two of these functions _prior to_ creating the `RelationalData` instance that holds all the metadata, but when we're restoring (denormalize) we can leverage the reldata instance.

src/gretel_trainer/relational/core.py
```diff
@@ -32,17 +40,57 @@ class ForeignKey:
...
     safe_ancestral_seed_columns: Optional[set[str]] = None
 
 
 class RelationalData:
     def __init__(self):
         self.graph = networkx.DiGraph()
+        self.relational_jsons: dict[str, RelationalJson] = {}
+
+    def restore(self, tableset: dict[str, pd.DataFrame]) -> dict[str, pd.DataFrame]:
+        """Restores a given a tableset (presumably output from some MultiTable workflow,
```
pimlock, line 66: nit: (or with "the"?)
```suggestion
        """Restores a given tableset (presumably output from some MultiTable workflow,
```

src/gretel_trainer/relational/core.py
```diff
@@ -180,17 +316,104 @@ def remove_foreign_key(self, table: str, constrained_columns: List[str]) -> None
...
+                            referred_columns=fk.parent_columns,
+                        )
+                else:
+                    metadata.data = data
+                    metadata.columns = set(data.columns)
+                    self._clear_safe_ancestral_seed_columns(table)
+            except KeyError:
+                raise MultiTableException(
+                    f"Unrecognized table name: {table}. If this is a new table to add, use `add_table`."
+                )
```
pimlock, line 361: nit: IMO this `try:except` could be a bit more local, with this much code in between, it's hard to understand where this error can be raised.

src/gretel_trainer/relational/backup.py
```diff
@@ -10,6 +10,7 @@
 @dataclass
 class BackupRelationalDataTable:
     primary_key: List[str]
+    invented_table_metadata: Optional[dict[str, str]] = None
```
pimlock, line 13: Is the word "invented" used here to potentially capture wider set of _inventions_ that lead to creating extra tables? For now can this be thought of as _nested_ tables?

pimlock: After more reading through this, it looks like a table is _invented_ if:
- it was flattened, or
- it was created from nested data

pimlock (commented):
Well, this PR is dense :)
It looks great! I got through most of it, added some nits and I think it would be easier to continue by walking through this together. I'm going to set up some time next week.

src/gretel_trainer/relational/backup.py
mikeknep: Ironically, I think your two bullet points could actually be two ways of describing a single situation—lifting the key-value pairs of a JSON dictionary to standalone columns sounds like "flattening nested data" to me 😄 

I feel like the word "invented" makes it clear that this is not a table that a user provided but one that was created internally; meanwhile, if we used "nested" I almost anticipate wondering "wait, is this the source table that does have nested data, or is it the table we created that has **de-nested** data?". Invented also feels a little future-proof in case we need to create more non-user-source tables for some reason.

src/gretel_trainer/relational/core.py
mikeknep: Good call, fixed

src/gretel_trainer/relational/backup.py
pimlock: Oh, I see, yeah, this makes sense, thanks for the explanation!

pimlock (approved):
Looks good! 

Thanks for adding some logging, I think we could add more `debug` over time (even around core relational), at least for me, it would make it easier to understand what's happening when I run it.

PR #109, opened 2023-05-17
Single transforms config

mikeknep:
Adds a new method, `train_transforms`, that accepts a single Transforms model config and applies it to all tables. This method accepts an optional list of tables passed as either `only` or `ignore` to limit which tables are trained. (These arguments match the optional list arguments on the Relational Trainer Connector's `extract` method.)

Users can include the name of a table as it appears in their db in the `only|ignore` argument and we will automatically extend that "request" to tables we invented from that source table (due to JSON). For example, if a `purchases` table had nested JSON and we ended up inventing 3 tables when we normalized that JSON data, the user can just pass the name `purchases` to explicitly include/exclude that table—they don't need to know the exact names of the tables we created under the hood.

This method will eventually fully replace `train_transform_models`, which is marked for deprecation here.

While here, updated a few of the type hints to the new style.

### DIFF
--- a/src/gretel_trainer/relational/multi_table.py
+++ b/src/gretel_trainer/relational/multi_table.py
@@ -33,6 +33,7 @@ from gretel_trainer.relational.core import (
     MultiTableException,
     RelationalData,
     Scope,
+    skip_table,
 )
 from gretel_trainer.relational.json import InventedTableMetadata, RelationalJson
 from gretel_trainer.relational.log import silent_logs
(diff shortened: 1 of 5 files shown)


src/gretel_trainer/relational/multi_table.py
```diff
@@ -557,7 +558,73 @@ def classify(self, config: GretelModelConfig, all_rows: bool = False) -> None:
...
+            )
+            self._transforms_train.models[table] = model
+
+        self._backup()
+
+        task = TransformsTrainTask(
+            transforms_train=self._transforms_train,
+            multitable=self,
+        )
+        run_task(task, self._extended_sdk)
```
mikeknep, line 595: This is copy/paste of the original method—I think I'll consolidate it into a single method. Also working on getting some unit tests around this, will push both to this branch.

tests/relational/test_train_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+    mt = MultiTable(ecom, project_display_name=tmpdir)
+    mt.train_transforms("transform/default")
+    transforms_train = mt._transforms_train
+
+    assert set(transforms_train.models.keys()) == set(ecom.list_all_tables())
+
+
+def test_train_transforms_only_includes_specified_tables(ecom, tmpdir, project):
+    mt = MultiTable(ecom, project_display_name=tmpdir)
+    mt.train_transforms("transform/default", only=["users"])
```
pimlock, line 41: Q: Is it possible to run `train_transforms` multiple times? For example I have 2 tables and want to use different config for each one:

```python
mt = MultiTable(ecom, project_display_name=tmpdir)
mt.train_transforms("transform/default", only=["users"])
mt.train_transforms("transform/redact-pii-nlp-off", only=["purchases"])
```

Would something like this work?

pimlock (approved):
Looks good!

tests/relational/test_train_transforms.py
mikeknep: Yes! If you called `mt.run_transforms()` after that snippet you wrote, you'll kick off jobs for both users and purchases. Also worth noting that if you called `train_transforms` twice and some table was included in both calls (e.g. first `only=["users"]` and then `only=["users", "purchases"]`), we'll essentially drop the first model on the floor in terms of MultiTable's internal state. (It will remain in the project, though.)

pimlock: Nice, thanks!

mikeknep: This felt worth capturing in a test

PR #111, opened 2023-05-22
Preserve tables before training

mikeknep:
### Main user-facing change

The `train` method is deprecated in favor of the new `train_synthetics` method. The latter accepts optional `only | ignore` lists to limit which tables are trained—the idea here is that if some table contains static reference data that you don't want synthesized, you can omit it from the synthetics workflow entirely.

The main reason for introducing this as a new method instead of optional arguments to the existing method is that the original `train` method resets the synthetic train state when called a second time. That felt like behavior a user _might_ rely on, so I figured it'd be safest to leave it alone and introduce a new method (which, FYI, works "additively"—see [this conversation](https://github.com/gretelai/trainer/pull/109#discussion_r1198129227) discussing it in the context of Transforms). I also think the original method name is too generic and it's nicer to specify "synthetics" in the method name given what all Relational can do.

Note: previously we let users skip table synthesis at `generate` time via the `preserve: list[str]` argument, but of course that method is called _after_ `train`/`train_synthetics` so we would have unnecessarily trained models for the preserved tables. The `preserve` argument still exists on `generate` and currently is _not_ marked for deprecation—we want to see if any customers have a use case for it (e.g. I have 10 tables, I know 2 will never be synthesized so I omit them from training, later during generation there is one more table I want to preserve with source data instead of synthesizing. TBD.)

### Delete models method

The same [conversation](https://github.com/gretelai/trainer/pull/109#discussion_r1198129227) with @pimlock on #109 got me thinking that we should have a `delete_models` method—it seems like a useful thing to have alongside the `only | ignore` params. It was a pretty straightforward thing to add, but we could also skip it / remove it from this PR. @gracecvking any thoughts?

### Internal improvements

- Previously we had been keeping track of which columns were used in model training so that the ancestral strategy could be sure not to include those columns in seed data. In #101 we moved the logic of which columns can be used in seeding to the core `RelationalData` class. Here, I realized since that info is calculated and cached there, we don't need to carry it in the MultiTable state and backups—we can just ask for it again—so `training_columns` is removed from several places.
- Previously we had a dedicated field related to synthetics generate state called `missing_model` for tables that didn't train successfully. However, since we now allow users to omit tables from training, we can no longer say "if we can't get a completed model, treat it as failed"—we now need to know if the table attempted training unsuccessfully, or was deliberately omitted. Fortunately, like above, we can calculate this from other data in state, so `missing_model` is dropped.
- We now create and upload the training data archive file _before_ submitting and waiting for the model training jobs

### DIFF
--- a/src/gretel_trainer/relational/multi_table.py
+++ b/src/gretel_trainer/relational/multi_table.py
@@ -816,7 +816,10 @@ class MultiTable:
         only: Optional[list[str]] = None,
         ignore: Optional[list[str]] = None,
     ) -> None:
-        """Train synthetic data models on each table in the relational dataset"""
+        """
+        Train synthetic data models for the tables in the tableset,
+        optionally scoped by either `only` or `ignore`.
+        """
         only, ignore = self._get_only_and_ignore(only, ignore)
 
         all_tables = self.relational_data.list_all_tables()
@@ -832,7 +835,7 @@ class MultiTable:
         # all its ancestors must also be omitted. In the future, it'd be nice to either find a way to
         # eliminate this requirement completely, or (less ideal) allow incremental training of tables,
         # e.g. train a few in one "batch", then a few more before generating (perhaps with some logs
-        # here informing the user of which required tables are missing).
+        # along the way informing the user of which required tables are missing).
         self._strategy.validate_preserved_tables(omit_tables, self.relational_data)
 
         training_data = self._prepare_training_data(include_tables)
--- a/src/gretel_trainer/relational/multi_table.py
+++ b/src/gretel_trainer/relational/multi_table.py
@@ -605,31 +605,12 @@ class MultiTable:
         only: Optional[list[str]] = None,
         ignore: Optional[list[str]] = None,
     ) -> None:
-        if only is not None and ignore is not None:
-            raise MultiTableException("Cannot specify both `only` and `ignore`.")
-
-        m_only = None
-        if only is not None:
-            m_only = []
-            for table in only:
-                m_names = self.relational_data.get_modelable_table_names(table)
-                if len(m_names) == 0:
-                    raise MultiTableException(f"Unrecognized table name: `{table}`")
-                m_only.extend(m_names)
-
-        m_ignore = None
-        if ignore is not None:
-            m_ignore = []
-            for table in ignore:
-                m_names = self.relational_data.get_modelable_table_names(table)
-                if len(m_names) == 0:
-                    raise MultiTableException(f"Unrecognized table name: `{table}`")
-                m_ignore.extend(m_names)
+        only, ignore = self._get_only_and_ignore(only, ignore)
 
         configs = {
             table: config
             for table in self.relational_data.list_all_tables()
-            if not skip_table(table, m_only, m_ignore)
+            if not skip_table(table, only, ignore)
         }
 
         self._setup_transforms_train_state(configs)
@@ -722,6 +703,30 @@ class MultiTable:
         self._backup()
         self.transform_output_tables = reshaped_tables
 
+    def _get_only_and_ignore(
+        self, only: Optional[list[str]], ignore: Optional[list[str]]
+    ) -> tuple[Optional[list[str]], Optional[list[str]]]:
+        if only is not None and ignore is not None:
+            raise MultiTableException("Cannot specify both `only` and `ignore`.")
+
+        only_and_ignore = []
+
+        for given_tables in [only, ignore]:
+            if given_tables is None:
+                only_and_ignore.append(None)
+                continue
+
+            modelable_tables = []
+            for table in given_tables:
+                m_names = self.relational_data.get_modelable_table_names(table)
+                if len(m_names) == 0:
+                    raise MultiTableException(f"Unrecognized table name: `{table}`")
+                modelable_tables.extend(m_names)
+
+            only_and_ignore.append(modelable_tables)
+
+        return tuple(only_and_ignore)
+
     def _prepare_training_data(self, tables: list[str]) -> dict[str, Path]:
         """
         Exports a copy of each table prepared for training by the configured strategy
@@ -780,6 +785,22 @@ class MultiTable:
         tables = self.relational_data.list_all_tables()
         self._synthetics_train = SyntheticsTrain()
 
+        training_data = self._prepare_training_data(tables)
+        self._train_synthetics_models(training_data)
+
+    def train_synthetics(
+        self,
+        *,
+        only: Optional[list[str]] = None,
+        ignore: Optional[list[str]] = None,
+    ) -> None:
+        """Train synthetic data models on each table in the relational dataset"""
+        only, ignore = self._get_only_and_ignore(only, ignore)
+        tables = [
+            table
+            for table in self.relational_data.list_all_tables()
+            if not skip_table(table, only, ignore)
+        ]
         training_data = self._prepare_training_data(tables)
         self._train_synthetics_models(training_data)
 
--- a/src/gretel_trainer/relational/multi_table.py
+++ b/src/gretel_trainer/relational/multi_table.py
@@ -251,9 +251,6 @@ class MultiTable:
             with tarfile.open(synthetics_training_archive_path, "r:gz") as tar:
                 tar.extractall(path=self._working_dir)
 
-        self._synthetics_train.training_columns = (
-            backup_synthetics_train.training_columns
-        )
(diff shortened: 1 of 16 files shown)


gracecvking (approved):
LGTM

src/gretel_trainer/relational/multi_table.py
```diff
@@ -767,22 +802,51 @@ def _train_synthetics_models(self, training_data: dict[str, Path]) -> None:
                     self._extended_sdk,
                 )
 
-        # TODO: consider moving this to before running the task
```
tylersbray, line 770: TODID

src/gretel_trainer/relational/multi_table.py
```diff
@@ -767,22 +802,51 @@ def _train_synthetics_models(self, training_data: dict[str, Path]) -> None:
...
-        self._artifact_collection.upload_synthetics_training_archive(
-            self._project, str(archive_path)
-        )
-
     def train(self) -> None:
-        """Train synthetic data models on each table in the relational dataset"""
+        """
+        DEPRECATED: Please use `train_synthetics` instead.
+        """
+        logger.warning(
```
tylersbray, line 809: Nice.

src/gretel_trainer/relational/multi_table.py
```diff
@@ -767,22 +802,51 @@ def _train_synthetics_models(self, training_data: dict[str, Path]) -> None:
...
+        only, ignore = self._get_only_and_ignore(only, ignore)
+
+        all_tables = self.relational_data.list_all_tables()
+        include_tables = []
+        omit_tables = []
+        for table in all_tables:
+            if skip_table(table, only, ignore):
+                omit_tables.append(table)
+            else:
+                include_tables.append(table)
```
tylersbray, line 838: Maybe I'm not at the part of this PR where it becomes clear... I see, opaquely here, that we use skip_table to add to the list of omit_tables.  But then it looks like we otherwise throw the table onto the include list.  Hoping it becomes clear somewhere else how we only have the only's here.  If it does, this is just a bookmark comment for me and can be safely ignored :wave:

Does the only filtering happen in list_all_tables?

src/gretel_trainer/relational/multi_table.py
```diff
@@ -767,22 +802,51 @@ def _train_synthetics_models(self, training_data: dict[str, Path]) -> None:
...
+
+        # TODO: Ancestral strategy requires that for each table omitted from synthetics ("preserved"),
```
tylersbray, line 840: If ignore drags in a tree of tables, does only also drag in a tree of tables?

tylersbray (approved):
I had questions but not concerns 🚢 

Oh, this PR had a long README but I didn't see a ton of docstrings.  And maybe as a bigger thing, is there a confluence page with an architecture/state diagram/strategy walkthrough...?

src/gretel_trainer/relational/multi_table.py
```diff
@@ -516,6 +510,35 @@ def _create_debug_summary(self) -> None:
...
+        only: Optional[list[str]] = None,
+        ignore: Optional[list[str]] = None,
+    ) -> None:
+        only, ignore = self._get_only_and_ignore(only, ignore)
+        tables = [
+            table
+            for table in self.relational_data.list_all_tables()
+            if not skip_table(table, only, ignore)
+        ]
+        if workflow == "classify":
```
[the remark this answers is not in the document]
mikeknep: This is the first time users pass a workflow "name" as an argument—until now it's all been method names (`classify`, `train_transforms`, etc.) or attributes (`synthetic_output_tables`, `transform_output_tables`). I'm almost tempted to continue that pattern here and change this from `delete_models(workflow: str, ...)` to three methods
```
delete_classify_models
delete_transform_models
delete_synthetics_models
```
Thoughts?

src/gretel_trainer/relational/multi_table.py
mikeknep: The [skip_table function](https://github.com/gretelai/trainer/blob/main/src/gretel_trainer/relational/core.py#L667-L676) got added in a separate branch and is being reused here. uses the `only` and `ignore` lists to determine if the provided `table` should be skipped or not.

(Should have a docstring.)

src/gretel_trainer/relational/multi_table.py
mikeknep: Currently the optional args are "strict"; if a user supplies either `only` or `ignore`, we try to perform exactly what they requested—i.e. either a) only run the tables specified in `only`, or b) run all tables except those specified in `ignore`. The call below this comment to the strategy will validate that the scope of tables is acceptable, and if it is not will raise an exception. We don't try to take on any logic like "if you excluded table Y you must have also intended to exclude table Z."

Does that + the other response above answer your question?

src/gretel_trainer/relational/multi_table.py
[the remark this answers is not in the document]
pimlock: Could this `delete_models` clean all of the state? Are there use cases where you'd want to clean up one, but not all of the workflows?

---

And a related question - I was trying to find a test that would train synthetics and transforms in the same `MultiTable` instance, but didn't find one. Is use case like that supported/makes sense?

src/gretel_trainer/relational/multi_table.py
```diff
@@ -722,15 +726,39 @@ def run_transforms(
         self._backup()
         self.transform_output_tables = reshaped_tables
 
+    def _get_only_and_ignore(
+        self, only: Optional[list[str]], ignore: Optional[list[str]]
+    ) -> tuple[Optional[list[str]], Optional[list[str]]]:
```
pimlock, line 731: nit: these table filters could be sets, right? As the order doesn't matter?

pimlock (approved):
LG!

src/gretel_trainer/relational/multi_table.py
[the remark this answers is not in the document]
mikeknep: > Are there use cases where you'd want to clean up one, but not all of the workflows?

In fairness, I'm not really aware of any specific use case for deleting either some or all models or workflows. I added this because it felt like an appropriate mechanism to have based on the (new/emerging) behavior of users being able to add tables to their set of trained tables, plus it being pretty easy to do.

Earlier, if you called `train` multiple times you would 1) train models for all tables both times, and 2) orphan the first set of models (they'd still exist in the project, but MultiTable would replace internal pointers to them with pointers to the second set of trained models).
```python
mt.train()
# mt._synthetics_train.models == {"users": "model_id_123", "events": "model_id_abc"}
mt.train()
# mt._synthetics_train.models == {"users": "model_id_789", "events": "model_id_xyz"}

# 4 models exist in the project
```

Now, depending on only/ignore args, you _could_ do something like this:
```python
mt.train_synthetics(only=["users"])
# mt._synthetics_train.models == {"users": "model_id_123"}
mt.train_synthetics(only=["events"])
# we didn't abandon the users model...
# mt._synthetics_train.models == {"users": "model_id_123", "events": "model_id_abc"}


# ...but if we call train again with users in scope, we'd abandon it like before
mt.train_synthetics(only=["users"])
# mt._synthetics_train.models == {"users": "model_id_789", "events": "model_id_abc"}
```
Something about this makes me feel like having a mechanism for deleting models should exist from a "library completeness" perspective. However, I'm open to the argument that this isn't a real requirement / the use case is too unclear at this point and that we should just drop this new `delete_models` method altogether / save it for later.

src/gretel_trainer/relational/multi_table.py
mikeknep: Yes for sure, and passing a particular table name multiple times makes no sense. I actually used lists figuring most users would find lists slightly more intuitive/familiar than sets, but hinting with set is probably more informative to anyone looking at the code / documentation, + if you pass a list anyways it'll still work since all we're doing is iterating over the collection. Pushing a commit for this.

src/gretel_trainer/relational/multi_table.py
tylersbray: Thanks for clarifying.

src/gretel_trainer/relational/multi_table.py
tylersbray: Yup, sounds good.  Thanks.

src/gretel_trainer/relational/multi_table.py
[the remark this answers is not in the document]
mikeknep: Update: I decided to remove this entirely for now, we can revisit if/when a well-defined use case for it arises.

PR #113, opened 2023-05-24
Fix bug related to empty normalized tables

mikeknep:
When parsing JSON, it's possible for all values to contain some empty list value—for example, in the unit tests `bball` fixture, each player has a `suspensions` key with empty list value. We want to be able to restore this field in the final denormalized/original-matching output data, but we **don't** want to submit an empty table for training.

We now include add empty invented tables to the graph (so we know where to put them back when denormalizing), but ensure they are not exposed to clients asking for MODELABLE tables.

As before this branch, we still do not support dataframes where _all columns_ contain JSON lists. Fortunately, as the unit test changes demonstrate, these changes have the added benefit of detecting that sooner and avoiding ending up in a bad state when this edge case happens.

### DIFF
--- a/src/gretel_trainer/relational/backup.py
+++ b/src/gretel_trainer/relational/backup.py
@@ -10,7 +10,7 @@ from gretel_trainer.relational.core import ForeignKey, RelationalData
 @dataclass
 class BackupRelationalDataTable:
     primary_key: list[str]
-    invented_table_metadata: Optional[dict[str, str]] = None
+    invented_table_metadata: Optional[dict] = None
 
 
 @dataclass
--- a/src/gretel_trainer/relational/backup.py
+++ b/src/gretel_trainer/relational/backup.py
@@ -36,7 +36,6 @@ class BackupRelationalJson:
     original_primary_key: list[str]
     original_columns: list[str]
     table_name_mappings: dict[str, str]
-    invented_table_names: list[str]
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


src/gretel_trainer/relational/backup.py
```diff
@@ -10,7 +10,7 @@
 @dataclass
 class BackupRelationalDataTable:
     primary_key: list[str]
-    invented_table_metadata: Optional[dict[str, str]] = None
+    invented_table_metadata: Optional[dict] = None
```
pimlock, line 13: Would it be worth keeping the first `str` and putting `Any` for value types?
I think we don't use any of these values directly, they are only used to populate `InventedTableMetadata`, right?

pimlock (approved):

src/gretel_trainer/relational/backup.py
mikeknep: Yeah good call. I feel like `dict` should _be_ `dict[str, Any]` instead of I guess effectively `dict[Any, Any]`.

You're correct that this dataclass is strictly capturing what goes in and out of the backup file, and the values are just used to populate the `InventedTableMetadata` object. I feel like it's useful to have a pretty strict definition of what to expect to see in the backup JSON files, but it does feel a little obnoxious at times since many structures are nearly or even exactly identical to the internal domain objects used more widely in the code.

PR #114, opened 2023-05-30
Notebook update

mikeknep:


mikeknep:
@gracecvking note: the `preserve_tables` argument is still supported and not even marked for deprecation, but it seems worth de-emphasizing in favor of scoping tables properly before training.

gracecvking (approved):

PR #118, opened 2023-06-03
Table extraction module + subsetting

johntmyers:
This PR introduces two new user-facing objects, a `ExtractionConfig` and a `TableExtractor` which are used to manage extracting tables from the configured target database. The current implementation is backwards compatible with the previous behavior, where entire tables are extracted by default. The existing params on `Connector.extract()` are marked for deprecation, and moving forward a user should exclusively use an `ExtractionConfig`.

The `TableExtractor` requires a storage directory as each table is extracted and stored as a CSV. For the current `Connector.extract()` update, we just create a temp directory to use as a means to get the DataFrames back into memory. One other note here is that we have to build up the `RelationalData` graph twice. The first time is because we need the "pure" SQL metadata in order to know what order to extract tables in and to support subsettings. Once we do the table extraction, we have to re-compute the `RelationalData` graph so it can incorporate the actual table contents in order to build up the invented JSON metadata. This is handled automatically by `TableExtractor` as an optional step when reading in tables.

This should put us on a path to have the tables themselves utilized from disk for future improvements to the actual model creation and transform/synthetics process downstream.


src/gretel_trainer/relational/connectors.py
```diff
@@ -41,51 +42,58 @@ def __init__(self, engine: Engine):
...
+                specific tables, ignore specific tables, and configure subsetting. Please
+                see the `ExtractorConfig` docs for more details.
+
+        NOTE: The `only`, `ignore`, and `schema` args will be deprecated. Only the `config` param
+        will be used in the future, which contains all of the params from above.
         """
+        if only or ignore or schema:
+            logger.warning(
+                "Deprecation warning: in the future only the `config` param will be suported, please use an `ExtractorConfig` object for this param."
+            )
```
mikeknep, line 78: Should we mark this entire `Connector#extract` method as deprecated and guide users towards using the `TableExtractor`? It seems like from an object responsibility perspective we're trying to simplify the Connector such that we'd eventually want users to basically implement a diff like the one below. Does this line up with what you're thinking?
```diff
connector = Connector.from_conn_string("...")
- rel_data = connector.extract()
+ config = ExtractorConfig(...)
+ extractor = TableExtractor(connector=connector, config=config, storage_dir=storage_dir)
+ extractor.sample_tables()  # *
+ rel_data = extractor.relational_data
```

\* I guess we'd currently need to set `refresh_relational_data=True` here, correct? Something about that feels strange but I'm not sure how I'd want to resolve it... have that value be the default? introduce some other function/method? something else?

src/gretel_trainer/relational/extractor.py
```diff
@@ -0,0 +1,518 @@
...
+
+    @property
+    def columns(self) -> list[str]:
+        return [column.name for column in self.table.columns]
+
+
+@dataclass
+class _PKValues:
+    """
+    Contains information that is needed to sample rows from an intermediate table
```
mikeknep, line 138: Is there a reason you use "intermediate" throughout and not "parent"? Intermediate makes me think that it's maybe some temporary/ephemeral table that gets "passed through" on the way to some more terminal table.

src/gretel_trainer/relational/extractor.py
```diff
@@ -0,0 +1,518 @@
...
+                values_list = chunk_df.to_records(index=False).tolist()
+                query = table_session.table.select().where(
+                    tuple_(
+                        *[table_session.table.c[col] for col in pk_values.column_names]
+                    ).in_(values_list)
+                )
+
+                with table_session.engine.connect() as conn:
+                    df_iter = pd.read_sql_query(query, conn, chunksize=self._chunk_size)
+                    write_count = _stream_df_to_path(df_iter, table_path)
```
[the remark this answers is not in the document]
johntmyers: Good question, let me look into seeing if the processing of each partition is sequential or not. If it's not, we could probably just use a `threading.Lock()` at the write step?

src/gretel_trainer/relational/extractor.py
johntmyers: Can't a table be both a parent and a child though? Maybe it's better to just say non-leaf.

src/gretel_trainer/relational/extractor.py
```diff
@@ -0,0 +1,518 @@
...
+
+
+@dataclass
+class _PKValues:
+    """
+    Contains information that is needed to sample rows from an intermediate table
+    """
+
+    table_name: str
+    values_ddf: dd.DataFrame  # pyright: ignore
```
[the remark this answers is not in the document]
johntmyers: Yes and I didn't really spend much time looking into it vs just having pyright stfu about up - seems very weird though - my guess it's based on how Dask is implemented/typed?

src/gretel_trainer/relational/extractor.py
mikeknep: Yeah... "non-leaf" might be least confusing here. FWIW though, you could describe the leaf tables as "non-parent" tables (because they don't have children i.e. tables with FKs pointing to them) that may or may not themselves be children. In other words, regardless of whether a table is a child or not, if it is a parent we'll be using _PKValues and if not we sample with the `SampleMode` enum.

src/gretel_trainer/relational/extractor.py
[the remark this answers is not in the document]
mikeknep: Yeah that's fair, can dig deeper into it later.

src/gretel_trainer/relational/connectors.py
johntmyers: I think we should just change the whole UX up once we have scaling working better (working only w/ table refs on disk). Thinking about it more I might just remove `refresh_relational_data` as a param, and push that actual logic back into `Connector.extract` - it doesn't seem like a useful param directly to users and if we want to move away from loading DFs into memory then it's kinda pointless.

I can also just remove the existing deprecation warning, and just implicitly pass the `extract` kwargs to the `ExtractorConfig` - we can just have customers use a custom config if they need to subset sooner.

mikeknep (approved):

src/gretel_trainer/relational/extractor.py
```diff
@@ -0,0 +1,518 @@
...
+                    tmp_ddf = tmp_ddf.rename(columns=rename_map)
+                    if values_ddf is None:
+                        values_ddf = tmp_ddf
+                    else:
+                        values_ddf = dd.concat([values_ddf, tmp_ddf])  # pyright: ignore
+
+        if parent_column_names:
+            values_ddf = values_ddf[  # pyright: ignore
+                parent_column_names
+            ].drop_duplicates()  # pyright: ignore
```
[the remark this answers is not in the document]
johntmyers: Yeah, good call. I think I had this before I added the `usecols` kwarg to `read_csv()`

PR #120, opened 2023-06-07
Custom synthetics configs

mikeknep:
With these changes, users are no longer limited to the synthetics blueprints configs and can provide fully customized model configs. You can supply one "main" config as well as table-specific configs if you want to get that granular:

```python
mt.train_synthetics(
    config=my_tabular_dp_config,
    table_specific_configs={"table_x": my_actgan_config},
)
```

~~Are these parameter values sufficiently clear? I didn't particularly like using the word "overrides" (e.g. `table_config_overrides`) because that feels to me like it'd be expecting "field-level" overrides on top of the `config` as a "base template", when in fact we are requiring _fully formed configs_ in both places.~~ EDIT: going with `table_specific_configs`

I recommend looking at the unit tests added to the diff for some behavior examples, but to call it out here in English: if you pass a `config` to the method call, it will overrule the blueprint set when you instantiate the MultiTable instance (the `gretel_model` attribute). We do still support `mt.train_synthetics()` for now, and in that case we continue to use that `gretel_model` blueprint config.

Looking ahead, we intend to deprecate `gretel_model` and require it be set when calling `train_synthetics` instead. We'll also move `strategy` as well. I'm not sure how soon we should add deprecation warnings; @gracecvking @mvansegbroeck any thoughts?

### DIFF
--- a/src/gretel_trainer/relational/model_config.py
+++ b/src/gretel_trainer/relational/model_config.py
@@ -1,5 +1,5 @@
 from copy import deepcopy
-from typing import Any
+from typing import Any, Optional
 
 from gretel_client.projects.models import read_model_config
 
@@ -8,6 +8,16 @@ from gretel_trainer.relational.core import (
     MultiTableException,
     RelationalData,
 )
+
+
+def get_model_key(config_dict: dict[str, Any]) -> Optional[str]:
+    try:
+        models = config_dict["models"]
+        assert isinstance(models, list)
+        assert isinstance(models[0], dict)
+        return list(models[0])[0]
+    except (AssertionError, IndexError, KeyError):
+        return None
 
 
 def _ingest(config: GretelModelConfig) -> dict[str, Any]:
--- a/src/gretel_trainer/relational/model_config.py
+++ b/src/gretel_trainer/relational/model_config.py
@@ -1,6 +1,7 @@
 from copy import deepcopy
 from typing import Any, Optional
 
+from gretel_client.projects.exceptions import ModelConfigError
 from gretel_client.projects.models import read_model_config
 
 from gretel_trainer.relational.core import (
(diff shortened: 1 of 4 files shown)


mvansegbroeck:
Thanks Mike. 

_>> Are these parameter values sufficiently clear?_ 
I don't have any better idea here w/o making the arguments name too long for the purpose of being more descriptive (such as main_config or dataset_config). I guess this looks good to me.

mvansegbroeck:
_>> deprecate gretel_model and require it be set when calling train_synthetics instead._
I would add a deprecation warning for now, but make sure to update all UCC notebooks so the deprecation warning becomes obsolete at some point and can be removed.

mvansegbroeck:
Q: What happens if a user specifies the ancestral strategy and at least one table is using a non-supported config?
Will it just fall back to independent for the entire datasets, or partially run ancestral strategy on the tables that are related to each other and are supporting ancestral?

mikeknep:
> I don't have any better idea here w/o making the arguments name too long for the purpose of being more descriptive

Same, every alternative I thought of seemed obnoxiously long. I think the names here are probably fine, but we'll see if anyone else has suggestions.

> What happens if a user specifies the ancestral strategy and at least one table is using a non-supported config?

All provided configs (`config` and everything in `table_configs`) must be compliant with the configured strategy; they are all checked prior to launching any jobs, and we raise an exception if any unsupported config is found.

gracecvking:
I also cant really think of a better parameter name... maybe "alt_configs" or "substitute_configs"? IDK. @sami-t might have a better parameter name idea. If not, I'm ok leaving it as table_configs

sami-t:
> I also cant really think of a better parameter name... maybe "alt_configs" or "substitute_configs"? IDK. @sami-t might have a better parameter name idea. If not, I'm ok leaving it as table_configs

I'm also fine with it. Is renaming `config` to `default_config` an option, or would that break backwards compatibility? At that point I think @mikeknep's concern around the `table_config_overrides` naming might be resolved. That could also remain as `table_configs`, or we could consider renaming to `alternate_configs` or even `alternative_table_configs` for maximum readability. I think the combination of default + override/alternative might make the behavior clearer.

tylersbray (approved):
LGTM.  Tests were super helpful for grokking it, thank you.

PR #129, opened 2023-06-20
Backup cols

mikeknep:
In the PR for column order (https://github.com/gretelai/trainer/pull/123), I added `columns` as a required field on the backup table dataclass, but we don't actually read that value back during the restore process. This PR drops that field so that backup files should be backwards-compatible a little while longer. (I believe we will need to add `columns` in the future, but hey, maybe not!).

Two other small changes here:
- When writing source data from CSV to DF (to upload in the `source_tables.tar.gz` archive), provide the columns argument explicitly. This probably doesn't have any effect in most cases, but is consistent with how we write our transforms and synthetics output files.
- The restore process was logging a deprecation warning about the `gretel_model` argument even though the user hadn't actually set that explicitly (it was coming from the backup file, again for backwards compatibility).

### DIFF
--- a/src/gretel_trainer/relational/backup.py
+++ b/src/gretel_trainer/relational/backup.py
@@ -10,7 +10,6 @@ from gretel_trainer.relational.core import ForeignKey, RelationalData
 @dataclass
 class BackupRelationalDataTable:
     primary_key: list[str]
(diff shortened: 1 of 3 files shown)


mckornfield (approved):
👌

PR #130, opened 2023-06-27
Source on disk

mikeknep:
Instead of storing tables' source data as Pandas DataFrames in memory, store them as CSV files on disk. Users can now provide either a dataframe or a string/Path to a CSV when calling `add_table`.

I did some initial profiling using the [Airline db from relational fit](https://relational.fit.cvut.cz/dataset/Airline), which self-reports as being 455MB. I used [tracemalloc](https://docs.python.org/3/library/tracemalloc.html) to measure both the "resting memory footprint" at certain checkpoints, as well as the peak memory size between those checkpoints. The table compares current released code on main / Trainer 0.9.0 with the code on this branch. 

Using the first row and main branch as an example, the way to read this table is:
> While running on `main` branch, after instantiating a Connector, the standing memory load was `1,132,735` bytes. While that method was running the memory spiked as high as `1,134,936`.

| Checkpoint | Main current | Main peak | Branch current | Branch peak |
| - | - | - | - | - |
| `conn = Connector.from_conn_str(...)` | 1,132,735 | 1,134,936 | 1,132,561 | 1,134,762 |
| `rd = conn.extract()` | 321,629,892 | 1,270,347,471 | 4,313,712 | 1,270,054,369 |
| `mt = MultiTable(rd)` | 322,222,376 | 923,687,442 | 4,902,519 | 5,140,588 |
| `mt.train_transforms(...)` | 323,666,151 | 628,013,664 | 6,285,388 | 6,328,929 |

Some observations:
- the general "standing pressure" is much lower since we're only storing pointers to files in memory when nothing else is actively going on
- largest memory spike is during extraction; no difference between this branch and main
- since transforms models train on unaltered source data, we don't need to load the data into memory at all (we now just pass along the file pointer as the data source), which is why that bottom-right corner stays so nice and small.

### DIFF
--- a/src/gretel_trainer/relational/core.py
+++ b/src/gretel_trainer/relational/core.py
@@ -15,6 +15,9 @@ Please see the specific docs for the "RelationalData" class on how to do this.
 from __future__ import annotations
 
 import logging
+import os
+import shutil
+import tempfile
 from dataclasses import dataclass, replace
 from enum import Enum
 from pathlib import Path
@@ -34,6 +37,9 @@ from gretel_trainer.relational.json import (
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_RELATIONAL_SOURCE_DIR = "relational_source"
+PREVIEW_ROW_COUNT = 5
+
 
 class MultiTableException(Exception):
     pass
@@ -50,6 +56,7 @@ class ForeignKey:
     parent_columns: list[str]
 
 
+UserFriendlyDataT = Union[pd.DataFrame, str, Path]
 UserFriendlyPrimaryKeyT = Optional[Union[str, list[str]]]
 
 
@@ -87,7 +94,7 @@ class Scope(str, Enum):
 @dataclass
 class TableMetadata:
     primary_key: list[str]
-    data: pd.DataFrame
+    source: Path
     columns: list[str]
     invented_table_metadata: Optional[InventedTableMetadata] = None
     producer_metadata: Optional[ProducerMetadata] = None
@@ -96,7 +103,7 @@ class TableMetadata:
 
 @dataclass
 class _RemovedTableMetadata:
-    data: pd.DataFrame
+    source: Path
     primary_key: list[str]
     fks_to_parents: list[ForeignKey]
     fks_from_children: list[ForeignKey]
@@ -118,7 +125,9 @@ class RelationalData:
     See the specific method docstrings for details on each method.
     """
 
-    def __init__(self):
+    def __init__(self, directory: Optional[Union[str, Path]] = None):
+        self.dir = Path(directory or DEFAULT_RELATIONAL_SOURCE_DIR)
+        os.makedirs(self.dir, exist_ok=True)
         self.graph = networkx.DiGraph()
 
     @property
@@ -176,7 +185,7 @@ class RelationalData:
         *,
         name: str,
         primary_key: UserFriendlyPrimaryKeyT,
-        data: pd.DataFrame,
+        data: UserFriendlyDataT,
     ) -> None:
         """
         Add a table. The primary key can be None (if one is not defined on the table),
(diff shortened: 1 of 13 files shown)


mikeknep:
Another larger runthrough with tracemalloc, using the (smaller) 10.3MB [SFScores](https://relational.fit.cvut.cz/dataset/SFScores) database:


Checkpoint | Main current | Main peak | Branch current | Branch peak
-- | -- | -- | -- | --
conn = Connector.from_conn_str(...) | 1,151,096 | 1,161,228 | 1,150,812 | 1,160,944
rd = conn.extract() | 8,315,849 | 26,839,407 | 2,827,917 | 26,779,260
mt = MultiTable(rd) | 8,909,101 | 15,948,522 | 3,412,425 | 3,699,777
mt.classify(...) | 9,074,523 | 14,618,511 | 3,569,290 | 4,063,625
mt.train_transforms(...) | 9,327,326 | 14,820,846 | 3,808,074 | 3,854,207
mt.run_transforms(...) | 14,409,295 | 23,160,397 | 8,947,244 | 14,633,737
mt.train_synthetics(...) | 18,806,941 | 19,540,049 | 13,326,966 | 14,741,821
mt.generate() | 31,055,466 | 63,558,885 | 25,620,783 | 62,624,282


The substantial increases after `run_transforms` and `generate` are almost surely due to the fact that we're currently stashing the final output data from those actions as dataframes on the MultiTable object (`mt.[transform|synthetic]_output_tables`). We have plans to deprecate that property and just expose pointers to the final output files on disk; will come in a separate PR.

tylersbray (approved):
I made it through all the changes... Looks like mostly adding the new param for the data source, updating the test fixtures etc.  It's nice that this change actually simplifies some of the dumping to csv that was already taking place and the data in your attached tables is pretty promising.  LGTM.

gracecvking (approved):
LGTM too

src/gretel_trainer/relational/core.py
```diff
@@ -118,7 +125,9 @@ class RelationalData:
     See the specific method docstrings for details on each method.
     """
 
-    def __init__(self):
+    def __init__(self, directory: Optional[Union[str, Path]] = None):
+        self.dir = Path(directory or DEFAULT_RELATIONAL_SOURCE_DIR)
+        os.makedirs(self.dir, exist_ok=True)
```
pimlock, line 130: Same option as above
```suggestion
        self.dir.mkdir(parents=True, exist_ok=True)
```

src/gretel_trainer/relational/json.py
```diff
@@ -129,6 +127,34 @@ def make_suffix(s):
...
+def jsonencode(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
+    def _jsonencode(x):
+        if isinstance(x, str):
+            return x
+        elif isinstance(x, (dict, list)):
+            return dumps(x)
+
+    copy = df.copy()
+    for col in cols:
+        copy[col] = copy[col].map(_jsonencode)
```
pimlock, line 139: Would it make sense to add a check for empty `cols` and returning the original df?

My thinking is that JSON cols won't be present in many datasets, and creating that is not needed in such cases (and in the context of memory consumption, maybe this could make some difference for large datasets?).

It does change the contract, as the function would return a copy in some cases and the original in others, so that's not great, but maybe worth considering, since this is quite internal detail and could be documented? Based on the current usage, it looks like this wouldn't cause any problems.

src/gretel_trainer/relational/json.py
```diff
@@ -129,6 +127,34 @@ def make_suffix(s):
     return hashlib.sha256(s.encode("utf-8")).hexdigest()[:10]
 
 
+def jsonencode(df: pd.DataFrame, cols: list[str]) -> pd.DataFrame:
```
pimlock, line 130: This is a great find on writing dataframes that contain dicts/lists as CSV and preserving its "JSON status".

tests/relational/conftest.py
```diff
@@ -147,7 +161,7 @@ def _setup_nba(
         referred_columns=["id"],
     )
 
-    return rel_data, states, cities, teams
+    yield rel_data, states, cities, teams
```
pimlock, line 164: I'm curious about using `yield` here instead of `return`. This extra `yield` is why you need to wrap it in `next()`, right?
Maybe that's necessary, e.g. if you need laziness of this, but I couldn't quite figure it out.

pimlock (approved):
Awesome work! Few minor nits/potential tweaks.

tests/relational/conftest.py
mikeknep: I think this is left over from some earlier machete work—I've changed it back to a regular `return` and it works fine, thanks for the catch!

src/gretel_trainer/relational/json.py
mikeknep: Nice idea, I've added it

PR #131, opened 2023-06-29
B2 backwards compatibility

mikeknep:
This PR includes changes to merge into the `b2` branch (not `main` yet!) that will ease users' transition from Benchmark v1 to v2 once we fully merge and release v2. It does not make that upgrade **fully** backwards compatible—there are still some breaking changes—but it should at least be a little smoother.

Since several people are already running Benchmark from the `b2` branch, I'll describe the changes in this PR from both perspectives. The sections below present the user-facing changes from `[section header]` to `[this branch]`; in other words, if we were to merge this PR _and_ merge b2 branch into main _and_ run a release of Trainer, these are the changes you'd have to address to start using the officially released Benchmark, depending on what you're using today.

### v1 / main branch / any Trainer version installed from pypi

- `Datatype.TABULAR_NUMERIC` and `Datatype.TABULAR_MIXED` are removed and will raise exceptions if you attempt to use them. You must change to `Datatype.TABULAR`.
- If you're passing a list of multiple sources to `make_dataset`, an exception will be raised.

(Everything else from v1 should work but start logging deprecation warnings where appropriate.)

### b2 branch

- Datatype enum variants should be capitalized, e.g. `Datatype.TABULAR` instead of `Datatype.tabular`. (I decided the change to lowercase variants on b2 branch was a mistake; capitalized enum variants are more pythonic.)
- Change `make_dataset` to `create_dataset`. (We want to keep the original signature of `make_dataset` and deprecate the function, instead of aggressively changing its signature completely.)

### DIFF
--- a/notebooks/benchmark.ipynb
+++ b/notebooks/benchmark.ipynb
@@ -74,7 +74,7 @@
    "metadata": {},
    "source": [
     "\n",
-    "When using your own data, indicate the datatype (select between: \"tabular_mixed\", \"tabular_numeric\", \"natural_language\", and \"time_series\"). Learn more in the [Benchmark docs](https://docs.gretel.ai/reference/benchmark#docs-internal-guid-31c7e29f-7fff-7936-54f8-737618a7e7f3).\n",
+    "When using your own data, indicate the datatype (select between: \"tabular\", \"natural_language\", and \"time_series\"). Learn more in the [Benchmark docs](https://docs.gretel.ai/reference/benchmark#docs-internal-guid-31c7e29f-7fff-7936-54f8-737618a7e7f3).\n",
     "\n",
     "Running in Google Colab? You can add your files to the Colab file system, and indicate the path like: \"/content/my_files/data.csv\""
    ]
@@ -85,7 +85,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# my_data = b.make_dataset(\"/PATH/TO/MY_DATASET.csv\", name=\"my-data\", datatype=\"INDICATE_DATATYPE\")"
+    "# my_data = b.create_dataset(\"/PATH/TO/MY_DATASET.csv\", name=\"my-data\", datatype=\"INDICATE_DATATYPE\")"
    ]
   },
   {
--- a/src/gretel_trainer/benchmark/custom/datasets.py
+++ b/src/gretel_trainer/benchmark/custom/datasets.py
@@ -1,11 +1,15 @@
+import logging
 import os
+import uuid
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Union
+from typing import Optional, Union
 
 import pandas as pd
 
 from gretel_trainer.benchmark.core import BenchmarkException, Datatype, get_data_shape
+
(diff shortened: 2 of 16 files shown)


src/gretel_trainer/benchmark/custom/datasets.py
```diff
@@ -57,3 +61,36 @@ def make_dataset(
...
+) -> CustomDataset:
+    logger.warning(
+        "`make_dataset` is deprecated and will be removed in a future release. Please use `create_dataset` instead."
+    )
+    if len(sources) > 1:
+        raise BenchmarkException(
+            "`make_dataset` no longer supports multiple sources. Please create separate datasets using `create_dataset`."
+        )
+
+    source = sources[0]
```
kboyd, line 81: nit: What do you think about being more defensive about the type here since python will happily let you call `make_dataset("myfile.csv")` and then you'll be very confused for a while about why it's looking for a file named `"m"`.

I know the type annotation says to call this with a list, do we think that's sufficient? Since string and dataframes will both happily return something from this `sources[0]` call and then the actual error won't show up until much farther into the code, being proactive to check for a list/iterable/tuple here before grabbing `sources[0]` feels useful to me. And it's another opportunity to direct them immediately to the non-deprecated `create_dataset`.

kboyd (approved):

src/gretel_trainer/benchmark/custom/datasets.py
mikeknep: Yeah this is a good call, I thought about this for a moment earlier and took the lazy route, but you're right, it's worth doing. I pushed a commit that adds a pretty basic check, LMK what you think.

notebooks/benchmark.ipynb
```diff
@@ -85,7 +85,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# my_data = b.make_dataset(\"/PATH/TO/MY_DATASET.csv\", name=\"my-data\", datatype=\"INDICATE_DATATYPE\")"
+    "# my_data = b.create_dataset(\"/PATH/TO/MY_DATASET.csv\", name=\"my-data\", datatype=\"INDICATE_DATATYPE\")"
```
lipikaramaswamy, line 88: Is it worth calling out here (or in the docs) that `make_dataset` has been deprecated?

lipikaramaswamy (approved):

notebooks/benchmark.ipynb
mikeknep: I'll definitely include it in the Trainer release notes when it we publish v2 ([examples](https://github.com/gretelai/trainer/releases)). Personally I would have the docs just present valid/current stuff and not mention in there (more confusing than good for new users), but I'll let @hhpang make the final call on that. (We will have some updates to make to the docs regardless.)

pimlock (approved):
Looks good!

notebooks/benchmark.ipynb
hhpang: oh man i'm so sorry i missed this message somehow. looks like we didn't call it out here, sgtm

PR #134, opened 2023-07-11
fix: unit test errors on new mac m2

benmccown:
A large portion of the unit tests were failing for me with a fresh clone of this repository on a brand new m2 macbook. Upon further digging I was able to identify sqlite3.connect as the issue. Specifically it wasn't honoring the uri params passed in the connection string and so it was creating a db on disk rather than an in memory shared db. This caused unit tests to throw errors.

ref: https://stackoverflow.com/questions/62502827/python-how-to-connect-sqlalchemy-to-existing-database-in-memory

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,15 @@ df = model.generate()
 
 ## Development
 
+Setup environment and install dependencies.
+
+```sh
+python -m venv venv
+source venv/bin/activate
+pip install -r requirements-dev.txt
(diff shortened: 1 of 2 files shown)


mckornfield (approved):
🆗

benmccown:
I went ahead and signed the CLA. Not sure how to trigger the `license/cla` bot to check again.

PR #135, opened 2023-07-12
fix: nested json table string length

benmccown:
In cases where we are parsing nested json with lengthy key strings the combination of key strings in the breadcrumb based table name string would result in an error when trying to train the model due to a maximum json filename length of 128 chars. Here we utilize the already existing make_suffix function (which ensures no sanitized_str collisions) and we truncate the length of the table name to ensure we do not exceed the 128 char limit.

### DIFF
--- a/src/gretel_trainer/relational/core.py
+++ b/src/gretel_trainer/relational/core.py
@@ -813,19 +813,14 @@ class RelationalData:
                 "foreign_keys": foreign_keys,
                 "is_invented_table": self._is_invented(table),
             }
-            if (
-                self.is_producer_of_invented_tables(table)
-                and (producer_metadata := self.get_producer_metadata(table)) is not None
-            ):
+            if (producer_metadata := self.get_producer_metadata(table)) is not None:
                 table_metadata["invented_table_details"] = {
                     "table_type": "producer",
                     "json_to_table_mappings": producer_metadata.table_name_mappings,
                 }
             elif (
-                self._is_invented(table)
-                and (invented_table_metadata := self.get_invented_table_metadata(table))
-                is not None
-            ):
+                invented_table_metadata := self.get_invented_table_metadata(table)
+            ) is not None:
                 table_metadata["invented_table_details"] = {
                     "table_type": "invented",
                     "json_breadcrumb_path": invented_table_metadata.json_breadcrumb_path,
--- a/src/gretel_trainer/relational/core.py
+++ b/src/gretel_trainer/relational/core.py
@@ -813,22 +813,23 @@ class RelationalData:
                 "foreign_keys": foreign_keys,
                 "is_invented_table": self._is_invented(table),
             }
-            if self.is_producer_of_invented_tables(table):
-                table_metadata["invented_table_details"] = {}
-                table_metadata["invented_table_details"]["table_type"] = "producer"
-                producer_metadata = self.get_producer_metadata(table)
-                if producer_metadata is not None:
-                    table_metadata["invented_table_details"][
-                        "json_to_table_mappings"
-                    ] = producer_metadata.table_name_mappings
-            elif self._is_invented(table):
-                table_metadata["invented_table_details"] = {}
-                table_metadata["invented_table_details"]["table_type"] = "invented"
-                invented_table_metadata = self.get_invented_table_metadata(table)
-                if invented_table_metadata is not None:
-                    table_metadata["invented_table_details"][
-                        "json_breadcrumb_path"
-                    ] = invented_table_metadata.json_breadcrumb_path
+            if (
+                self.is_producer_of_invented_tables(table)
+                and (producer_metadata := self.get_producer_metadata(table)) is not None
+            ):
+                table_metadata["invented_table_details"] = {
+                    "table_type": "producer",
+                    "json_to_table_mappings": producer_metadata.table_name_mappings,
+                }
+            elif (
+                self._is_invented(table)
+                and (invented_table_metadata := self.get_invented_table_metadata(table))
+                is not None
+            ):
+                table_metadata["invented_table_details"] = {
+                    "table_type": "invented",
+                    "json_breadcrumb_path": invented_table_metadata.json_breadcrumb_path,
+                }
             tables[table] = table_metadata
 
         return {
--- a/src/gretel_trainer/relational/core.py
+++ b/src/gretel_trainer/relational/core.py
@@ -806,13 +806,31 @@ class RelationalData:
                         "parent_columns": key.parent_columns,
                     }
                 )
-            tables[table] = {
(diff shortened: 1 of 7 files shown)


benmccown:
Let me know if I am missing anything in the unit test coverage. I think the tests I added should cover us but I wasn't sure if there were any other existing tests that would benefit from having the real `sanitize_str` method included in their run instead of the mocked function.

benmccown:
The latest commits incorporate feedback gathered via slack to name invented tables `<producer_table>_invented_<uuid>`. We also stored the resulting mapping info in the debug_summary file now.

tests/relational/test_relational_data_with_json.py
```diff
@@ -192,47 +220,74 @@ def test_invented_json_column_names(documents, bball):
...
-    assert documents.get_primary_key("purchases-sfx") == ["id", "~PRIMARY_KEY_ID~"]
+    assert documents.get_primary_key(purchases_root_invented_table) == [
+        "id",
+        "~PRIMARY_KEY_ID~",
+    ]
 
     # Setting an existing primary key to None puts us in the correct state
     assert len(documents.list_all_tables(Scope.ALL)) == 5
-    original_payments_fks = documents.get_foreign_keys("payments")
+    original_payments_fks: list[ForeignKey] = documents.get_foreign_keys("payments")
```
mikeknep, line 238: Just curious if the explicit type hint was needed here?

tests/relational/test_relational_data_with_json.py
```diff
@@ -192,47 +220,74 @@ def test_invented_json_column_names(documents, bball):
     ]
 
 
-def test_set_some_primary_key_to_none(documents, bball):
+@patch("gretel_trainer.relational.json.make_suffix")
+def test_set_some_primary_key_to_none(mock_make_suffix, documents):
+    # We need to reset the side_effect iterator later in the test, so we manually mock it
+    mock_make_suffix.side_effect = [f"sfx{i}" for i in range(1, 50)]
```
mikeknep, line 226: I think if you update the conftest implementation to yield the mock object (instead of just yielding nothing), you can avoid having to patch each of these tests and (re)define the initial implementation at the top of each of them. You'll still need to reset the side effect at the appropriate time (right before calling `set_primary_key`), of course. Something like this (I'm short-handing the fixture a bit, this example doesn't have your new changes)

```python
# conftest
def static_suffix():
    with patch("gretel_trainer.relational.json.make_suffix") as make_suffix:
        make_suffix.side_effect = itertools.count
        yield make_suffix  # instead of just `yield` here


# these tests
def test_set_some_primary_key_to_none(static_suffix, documents):
    # setup stuff, should take advantage of the autouse fixture from conftest

    # but now we have a handle to it and can reset it here
    static_suffix.side_effect = itertools.count
    documents.set_primary_key(...)

**Review by mikeknep — COMMENTED** [review]
Awesome work! Most of these suggestions come down to style / lesser-known features and functions; there is one test change that is confusing me.

File: tests/relational/test_relational_data_with_json.py [hunk 3]
```diff
@@ -175,15 +202,16 @@ def test_invented_json_column_names(documents, bball):
...
         "~PRIMARY_KEY_ID~",
         "purchases~id",
         "content",
         "array~order",
     ]
 
     # If the source table does not have a primary key defined, one is created on the root invented table
-    assert bball.get_table_columns("bball-sfx") == [
+    # (bball_teams_invented_table is the root invented table)
+    assert bball.get_table_columns(bball_teams_invented_table) == [
```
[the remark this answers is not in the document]
benmccown: This is a great question and is a callout I didn't catch. The issue is that the purchases tables call the mocked suffix method first and consume `sfx1` and `sfx2` for the purchases tables, so the bball root invented table becomes `bball_invented_sfx3`. When iterating on my code I first hard coded that string in this spot and then did a search and replace for the string with the named bball table vars. So the test/code is correct but the variable naming makes it seem like there's a logical error.

The thing is, with both fixtures being set up before any of this test runs, to properly set up variable naming we'd need to reset the make_suffix mock counter and then reingest tables for bball to fix the strings. Going to give some thought to the best way to do this.

tests/relational/test_relational_data_with_json.py
benmccown: For me the conftest fixtures that yield RelationalData objects are not showing type hints correctly in VS Code.

image

Probably the better way would be to type hint `documents` to RelationalData in the input params of the test, but maybe there's something we could to to fix this universally for all tests without needing type hints? Not sure.

tests/relational/test_relational_data_with_json.py
benmccown: Makes sense, cooking up a revision.

tests/relational/test_relational_data_with_json.py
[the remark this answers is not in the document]
mikeknep: Oof, yeah that's subtle. Would splitting this test into two separate tests that each only use one of the fixtures solve this conflict?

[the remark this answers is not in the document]
benmccown: Yes, I think that would simple and direct. I'll do that.

tests/relational/test_relational_data_with_json.py
mikeknep: Ooh interesting. So, I used to have type hints on the various RelationalData fixtures in conftest, but I removed them in a [recent PR](https://github.com/gretelai/trainer/pull/130) (here's the [file in the full commit](https://github.com/gretelai/trainer/commit/2b23892796c7a825ebd931b4953110939db5c977#diff-59ef55cb9903ff34cbd594a84ce5791847e83955caeac3361d3f6f9d29024006))—switching from `return` to `yield` was leading to some confusing errors in pyright and I think I just punted on it 😞 

My hypothesis is if we can restore type hints to those fixtures (however they need to be adjusted for the return/yield change), this problem will be solved throughout the rest of the tests.

src/gretel_trainer/relational/core.py
```diff
@@ -806,13 +806,32 @@ def debug_summary(self) -> dict[str, Any]:
...
+            ):
+                table_metadata["invented_table_details"] = {
+                    "table_type": "producer",
+                    "json_to_table_mappings": producer_metadata.table_name_mappings,
+                }
+            elif (
+                self._is_invented(table)
+                and (invented_table_metadata := self.get_invented_table_metadata(table))
+                is not None
+            ):
```
mikeknep, line 828: You can remove the "first half" of both these conditionals, since they are essentially helper methods that under the hood perform the same check that the "second half" of each conditional check does (and maps it to a boolean).

mikeknep (approved):
Great work!

PR #139, opened 2023-07-21
Introduce Session as a job executor to support use cases other than compare.

pimlock:
Introduces `Session` as an encapsulation of _given a list of jobs, run them all and give me results_.
And uses that for 2 use cases:
- compare - given 2 lists: datasets and models, create cross product of them and run all the jobs.
- launch - given list of (dataset, model), run them all.

### DIFF
--- a/src/gretel_trainer/benchmark/comparison.py
+++ b/src/gretel_trainer/benchmark/comparison.py
@@ -1,36 +1,21 @@
 from __future__ import annotations
 
 import logging
-from concurrent.futures import Future, ThreadPoolExecutor
 from inspect import isclass
 from pathlib import Path
-from typing import Any, Optional, Type, Union, cast
+from typing import Optional, Type, Union, cast
 
 import pandas as pd
 from gretel_client import configure_session
-from gretel_client.helpers import poll
-from gretel_client.projects import Project, create_project, search_projects
-from gretel_client.projects.jobs import Job
 
 from gretel_trainer.benchmark.core import BenchmarkConfig, BenchmarkException, Dataset
 from gretel_trainer.benchmark.custom.datasets import CustomDataset
 from gretel_trainer.benchmark.custom.models import CustomModel
-from gretel_trainer.benchmark.custom.strategy import CustomStrategy
-from gretel_trainer.benchmark.executor import Executor
 from gretel_trainer.benchmark.gretel.datasets import GretelDataset
-from gretel_trainer.benchmark.gretel.models import GretelAuto, GretelModel
-from gretel_trainer.benchmark.gretel.strategy_sdk import GretelSDKStrategy
-from gretel_trainer.benchmark.gretel.strategy_trainer import GretelTrainerStrategy
+from gretel_trainer.benchmark.gretel.models import GretelModel
+from gretel_trainer.benchmark.session import JobSpec, Session, launch, model_name
 
 logger = logging.getLogger(__name__)
-
-ALL_REPORT_SCORES = {
-    "SQS": "synthetic_data_quality_score",
-    "FCS": "field_correlation_stability",
-    "PCS": "principal_component_stability",
-    "FDS": "field_distribution_stability",
-    "PPL": "privacy_protection_level",
-}
 
 
 DatasetTypes = Union[CustomDataset, GretelDataset]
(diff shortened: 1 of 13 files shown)


src/gretel_trainer/benchmark/comparison.py
```diff
@@ -42,304 +27,37 @@
...
-    def export_results(self, destination: str) -> None:
-        self.results.to_csv(destination, index=False)
+    model_instances = _create_models(models)
+    _validate_compare(model_instances, datasets)
 
-    def wait(self) -> Comparison:
-        [future.result() for future in self.futures.values()]
-        return self
+    # TODO(pm): I think moving this outside of Session makes sense, so it can be configured externally
+    configure_session(api_key="prompt", cache="yes", validate=True)
```
mikeknep, line 42: You mean outside of any Benchmark library code entirely, right? Yes, I agree—in fact, we made that change for [Trainer](https://github.com/gretelai/trainer/pull/86) and [Relational](https://github.com/gretelai/trainer/pull/87) already to support running in hybrid environments; I think I just missed Benchmark due to the nature of the long-running `b2` branch.

tests/benchmark/test_bad_setup.py
```diff
@@ -21,32 +22,36 @@ class SomeException(Exception):
                 datasets=[iris],
                 models=[GretelLSTM],
                 config=BenchmarkConfig(
-                    working_dir="should_not_be_created",
+                    working_dir=should_not_be_created,
                 ),
             )
 
-    assert not os.path.exists("should_not_be_created")
+    assert not should_not_be_created.exists()
```
mikeknep, line 29: I love your commitment to `pathlib` over `os` 😄 👍

src/gretel_trainer/benchmark/session.py
```diff
@@ -0,0 +1,374 @@
...
+    # TODO(pm): for simplicity just use Dataset instead of union
+    dataset: Dataset
+
+    # TODO(pm): for simplicity, we only allow storing instantiated models here, no Type[X]
+    model: Union[GretelModel, CustomModel]
+
+
+def launch(*, jobs: list[JobSpec], config: Optional[BenchmarkConfig] = None) -> Session:
+    session = Session(jobs=jobs, config=config)
+    return session.prepare().execute()
```
mikeknep, line 70: I definitely agree that `JobSpec` is a lot simpler when the types of the two fields are constrained/standardized to what you have defined here, and I like `Session` accepting a list of those objects to use internally with `compare` handling the process of creating that list. However, I think `launch` should accept a list of some different dataclass that, like `compare`, uses the wider types (`DatasetTypes`, `ModelTypes`).

Right now we document two kinds of datasets: Gretel-curated and create-your-own. I don't really want to have to explain "there's also a third, generic version of Datasets that both those two can be coerced to and that we really use under the hood...". In other words, I want to keep the API surface area smaller. Similarly, with models like `GretelACTGAN` (which is a class name, but in current usage kind of feels more like a constant), I think it'd be odd to have to explain "you can pass `GretelACTGAN` to `compare` but you can't use it as a model in a `JobSpec` passed to `launch`."

This implementation of `launch` doesn't really add a _whole_ lot over just manually instantiating a Session and calling prepare + execute. It feels to me like if someone wants to go down to a lower level, they should do exactly that (init+prepare+execute), but if they want to operate at a more convenient higher level they should use `compare` or `launch` and be able to pass nearly-identical arguments to those two functions. It kind of seems like instead of two "levels" of API abstraction (high = `compare/launch`, low = `Session(...).prepare().execute()`), there are three here, with `launch` being "somewhere in the middle" between high and low.

src/gretel_trainer/benchmark/session.py
```diff
@@ -0,0 +1,374 @@
...
+    source: str, project: Project, trainer: bool
+) -> Optional[str]:
+    if trainer:
+        return None
+
+    return project.upload_artifact(source)
+
+
+def _make_run_key(model: Union[GretelModel, CustomModel], dataset: Dataset) -> RunKey:
+    return RunKey((model_name(model), dataset.name))
```
mikeknep, line 374: This is a bit outside the scope of this PR, and I might be missing something, but I noticed it seems like `RunKey` could now just be a property of the `JobSpec` dataclass instead of needing this builder function.

tests/benchmark/test_bad_setup.py
pimlock: Haha, yeah :D
It's interesting though, with Python's dynamic typing, anytime I need to do `something.foo()` it makes me a bit nervous, as the `something` can be anything, but `os.path.exists("foo")` feels safer :)

src/gretel_trainer/benchmark/session.py
pimlock: > However, I think launch should accept a list of some different dataclass that, like compare, uses the wider types (DatasetTypes, ModelTypes).

I agree. For now I wanted to do "can I extract this thing and make tests pass" and another iteration would be for the cleanup, so thanks for all this feedback. That `launch` in the current version is just a placeholder. I agree that `launch` and `compare` should be symmetrical as much as possible.

The main reason why I didn't implement `launch` properly was a question of what it should accept. I think the best option would be to have more permissive `JobSpec`, but I wasn't sure about introducing another class just for that. Then it could be a list of tuples or 2 lists, which would have to be same length.

From the point of view of similarity with `compare`, I'd be leaning towards list of tuples I think. What do you think?

src/gretel_trainer/benchmark/session.py
pimlock: Good idea!

src/gretel_trainer/benchmark/comparison.py
pimlock: Yes, that was my thinking.
Should I remove it then? Or leave as-is to be handled in a separate PR?

src/gretel_trainer/benchmark/session.py
mikeknep: OK yeah that "order of operations" on this branch totally makes sense, glad we're on the same page.

I think list of tuples is good. I definitely like that better than the two lists option. Compared to a new more permissive dataclass, my gut reaction was concern over "what if the user gets the order wrong", i.e. `(dataset, model)` vs. `(model, dataset)`, but thinking about it some more, if that happens I think it's fine to just raise whatever runtime error occurs, since we do all our "prep work" in advance before kicking off any jobs.

src/gretel_trainer/benchmark/comparison.py
mikeknep: You can go ahead and take it out now.

src/gretel_trainer/benchmark/session.py
```diff
@@ -0,0 +1,346 @@
...
+            return "Finished"
+        elif future.cancelled():
+            return "Cancelled"
+        else:
+            return "Running"
+
+    # TODO(pm): thoughts on renaming this to e.g. "_create_gretel_strategy()"?
+    #  From the name, I'd expect something to be set on the `self`, but it's returned,
+    #  so it's a bit confusing. Unless "set" means something else here.
+    def _set_gretel_strategy(
```
mikeknep, line 259: 👍 you're right, "create" is a better verb here

src/gretel_trainer/benchmark/session.py
```diff
@@ -0,0 +1,346 @@
...
+        trainer_project_index: int,
+        artifact_key: Optional[str],
+    ) -> Union[GretelSDKStrategy, GretelTrainerStrategy]:
+        # TODO(pm): this is to make typechecker happy, couldn't figure out a nicer way
+        #  to do this, without much refactoring.
+        #  Why I want to have job here? Because I needed to pass additional context
+        if not isinstance(job.model, GretelModel):
+            raise BenchmarkException(
+                "_setup_gretel_run can only be used with GretelModel!"
+            )
```
mikeknep, line 272: I think we would have to make `JobSpec` generic with respect to model type, and then for example constrain this method to take a `JobSpec[GretelModel]`. Been awhile since I've played with generics in python typehints, not sure how much effort would be required / if its worth it 🤷

src/gretel_trainer/benchmark/job_spec.py
```diff
@@ -0,0 +1,50 @@
+from dataclasses import dataclass
+from typing import Union
+
+from gretel_trainer.benchmark.core import Dataset
+from gretel_trainer.benchmark.custom.models import CustomModel
+from gretel_trainer.benchmark.gretel.models import GretelModel
+
+
+# TODO(pm): Move this and JobSpec to `core`?
```
mikeknep, line 9: Seems reasonable to me if you want; I don't have a strong opinion

src/gretel_trainer/benchmark/session.py
pimlock: Good idea, I was thinking about that, but didn't know if that would be a path worth taking. With your comment I went into the rabbit whole and this is what I got to, which works relatively well:
- it fails type check in places where I'd expect it to and requires `cast` (it cannot figure things out on its own, as `isinstance` cannot be used with generic type param)
- is reasonably easy to reason about/read/write

```python
AnyModelType = Union[GretelModel, CustomModel]
MODEL_BASE_TYPE = TypeVar('MODEL_BASE_TYPE', bound=AnyModelType, covariant=True)
# covariant=True, so that `list[GretelModel]` can be used where `list[AnyModelType]` is used

@dataclass
class JobSpec(Generic[MODEL_BASE_TYPE]):
    dataset: Dataset
    model: MODEL_BASE_TYPE
    
    def make_run_key(self):
        return RunKey((model_name(self.model), self.dataset.name))


class RandomModel(CustomModel):
    def train(self, source: Dataset, **kwargs) -> None:
        ...

    def generate(self, **kwargs) -> pd.DataFrame:
        ...

def foo(job: JobSpec[GretelModel]):
    print(job.model.name)

def bar(jobs: list[JobSpec[AnyModelType]]):
    for job in jobs:
		foo(job) # fails type check
        if isinstance(job.model, GretelModel):
			foo(job) # fails type check
            foo(cast(JobSpec[GretelModel], job)) # ok
			# note: that `cast` is needed here, as there is no way to "isinstance" based on the generic type, and the `isinstance` on `job.model` doesn't translate to job being `JobSpec[GretelModel]`.

if __name__ == '__main__':
    job = JobSpec(Dataset("a"), RandomModel())
    bar([job]) # ok, list[JobSpec[RandomModel]] can be used where list[JobSpec[AnyModelType]] is required, because of `covariant=True`
    foo(job) # fails type check
```

What do you think? :)

src/gretel_trainer/benchmark/job_spec.py
pimlock: Thanks, I'm going to leave it separate for now then.

src/gretel_trainer/benchmark/comparison.py
```diff
@@ -42,304 +26,52 @@
...
-        executor = Executor(
-            strategy=strategy,
-            run_identifier=run_identifier,
-            evaluate_project=self._project,
-            config=self.config,
+def launch(
+    *,
+    jobs: list[tuple[DatasetTypes, ModelTypes]],
+    config: Optional[BenchmarkConfig] = None,
+) -> Session:
```
pimlock, line 58: I put `launch` in here for now, as `compare` and `launch` are using the same helpers.

It doesn't fit 100%, so I was thinking of extracting that to somewhere else, since `gretel_trainer.benchmark.comparison.compare` is documented/main interface, I think renaming that module wouldn't be a good idea.

mikeknep: I think we've only documented importing `compare` directly from `gretel_trainer.benchmark`. In v2 that import is basically just an alias ([see here](https://github.com/gretelai/trainer/blob/b2/src/gretel_trainer/benchmark/__init__.py#L3)). In v1, the __init__ file [defines the public compare function](https://github.com/gretelai/trainer/blob/main/src/gretel_trainer/benchmark/__init__.py#L61-L67) and translates arguments + defaults to a config that gets passed to an "internal" function signature (in hindsight I'm not a fan of this and am glad it has changed).

So I think it's fair game to rename this module. First idea that comes to my head is `entrypoints.py`, maybe?

src/gretel_trainer/benchmark/session.py
pimlock: Very cool, I didn't know about `TypeGuard`, but it looks like exactly what's needed to solve this. Good to know we could use it with `typing_extensions`, even with our Python version.

Evaluating both options - `cast` vs `TypeGuard`, since the usage is quite limited, we'd needed it in a single spot right now, I'm wondering if adding extra dependency to have `TypeGuard` is worth in this case. 
My initial thinking is to just keep the `cast`, but since it's a small library without any dependencies, we could also just do that. Are there any other spots it could be useful?

src/gretel_trainer/benchmark/comparison.py
pimlock: Awesome, that's great you had these aliases in the main module, that makes it easier to move things around.
`entrypoints.py` sounds great!

src/gretel_trainer/benchmark/session.py
mikeknep: 1. Potentially could be useful [here](https://github.com/gretelai/trainer/blob/b2/src/gretel_trainer/benchmark/comparison.py#L94-L101)? The current code block isn't that bad, but I'd be curious if TypeGuard could clean it up a little.
2. A while back I [pinned pyright](https://github.com/gretelai/trainer/pull/105) when CI unexpectedly pulled a new release that caught many more type warnings that I didn't have time to address. I remember one of the main categories of warnings had to do with the difference between `Job` and `Model | RecordHandler`. I'm not certain of this, but I suspect we might want to ditch using the client `Job` class in our type hints and replace it with a generic that gets narrowed as needed via TypeGuard to either of the two subclasses (model/RH).

🤷

src/gretel_trainer/benchmark/session.py
```diff
@@ -32,136 +29,75 @@
...
+                data_source_map[job.dataset.data_source] = artifact_key
+
+            if is_gretel_model(job):
+                self._setup_gretel_run(job, artifact_key, _trainer_project_index)
                 _trainer_project_index += 1
 
-            for model in self.custom_models:
-                self._setup_custom_run(dataset, model, artifact_key)
+            else:
+                assert is_custom_model(job)
```
pimlock, line 90: This assert is needed to make the next line happy. I initially had `elif`, but I think this is better, as there are no other cases.
There are some discussions on type narrowing in the negative case [here](https://github.com/python/typing/discussions/1013), but there is nothing yet that could be used.

mikeknep: Interesting read in that discussion, thanks for linking to it! I think I would've landed on using `elif` here personally, but I don't feel too strongly about it. Obviously we shouldn't ever hit this, but I suppose if this assertion ever does explode at runtime we'll have 1) an extremely unusual/unexpected edge case on our hands, and 2) hopefully a better lead on resolving the situation than whatever the error message would end up being otherwise?

Is it worth a comment in the code?

src/gretel_trainer/benchmark/entrypoints.py
```diff
@@ -0,0 +1,137 @@
...
+            JobSpec(
+                dataset=_standardize_dataset(dataset, config.working_dir),
+                model=_create_model(model),
+            )
+            for dataset, model in jobs
+        ]
+    except Exception as e:
+        if working_dir_cleanup is not None:
+            working_dir_cleanup()
+        raise e
```
pimlock, line 92: @mikeknep based on tests in `test_bad_setup.py` I arrived with this, my thinking was this:
- how test describes this case is "if we weren't able to run anything, i.e. validation failed or client is not configured, we don't want to create the working dir"
- I think that working dir not existing was a proxy for checking that we didn't do any work, I wasn't sure if that dir not existing was an important part

At the end, as this is right now, the contract for that working dir is this:
- if it existed, we leave it, even with new files that we might have written as part of `_standarize_dataset`
- if it didn't exists, we will remove it if we get an exception before any work was done (Q: should this include failures in `prepare()`?)
- regardless if it existed or not, we will leave it if we started some work, the thinking here is that leaving these files may be useful for any troubleshooting/investigations.

One last someone related question: `Session` has `cleanup`, which removes the project that was created, should working dir be included in that cleanup?

src/gretel_trainer/benchmark/session.py
pimlock: Good call. I think between adding a comment and switching it to `elif` + `else` with exception, which acts as a comment the latter sounds like a better option, so I changed it to that.

src/gretel_trainer/benchmark/session.py
pimlock: Thanks! Added `typing-extensions` and using `TypeGuard` in that spot now.

src/gretel_trainer/benchmark/entrypoints.py
mikeknep: I don't think I had really ever thought this through so thoroughly :joy: I don't have a particularly strong opinion here, nor do I think we've documented/committed to any specific behavior. My main thought is that since this is often running on someone's local machine, we shouldn't delete anything that may have existed before we started working, so your first bullet point about the contract seems most important to me.

RE: `Session#cleanup`, I think that was primarily to assist with cleaning up _multiple_ projects, which happens when running with Trainer enabled—in that mode, each run gets its own project because datasets can be broken into smaller partitions and Trainer has its own artifact eviction strategy. I'm inclined to say that it should continue only doing that (i.e. a helper over gretel client SDK/API calls) and not do anything with the working directory—leave that to users explicitly doing an `rm -rf`. WDYT?

src/gretel_trainer/benchmark/entrypoints.py
```diff
@@ -0,0 +1,129 @@
...
+        ]
+        if validate is not None:
+            validate(job_specs)
+
+        if any(is_gretel_model(job) for job in job_specs):
+            _verify_client_config()
+
+    except Exception as e:
+        if working_dir_cleanup is not None:
+            working_dir_cleanup()
```
mikeknep, line 83: tiny style nit: you could initially set `working_dir_cleanup = lambda: pass` and then just always call it here without checking for `None`.

mikeknep: Or, hm, I guess `lambda: pass` is invalid, but `lambda: None` works.

src/gretel_trainer/benchmark/executor.py
```diff
@@ -102,7 +103,7 @@ def _train(self) -> None:
             self.strategy.train()
             self._log("training completed successfully")
         except Exception as e:
-            self._log("training failed")
+            self._log("training failed", level=logging.ERROR, exc_info=e)
```
mikeknep, line 106: So simple, yet so useful! 😍

src/gretel_trainer/benchmark/entrypoints.py
pimlock: Great suggestion!

src/gretel_trainer/benchmark/entrypoints.py
pimlock: Sounds good on the `Session#cleanup`.

And on the working dir itself - main reason I went into that rabbit hole was that the test was using that dir to assert that no work was done and I was thinking of using a different "signal" for that, as the working dir creation seemed like a low level detail that may not be the best thing to "inspect" in the test, but then started wondering if that dir's existence is actually part of the contract and took this into that direction :).

TL;DR for the handling as it is right now:
- if dir was there, leave it there with any contents that were there previously or created during the initial setup
- if dir wasn't there and we didn't get to do any work (which is indicated by running stuff on `Session`), then remove it, so it doesn't litter someone's disk.
- if dir wasn't there, but we did work, leave it.

mikeknep: Seems reasonable to me 👍

src/gretel_trainer/benchmark/job_spec.py
```diff
@@ -0,0 +1,65 @@
...
+    model: MODEL_BASE_TYPE
+    """
+    Instance of a model that will be used.
+    """
+
+    metadata: Optional[dict[str, Any]] = None
+    """
+    Arbitrary metadata that can be attached to the spec for context that may be used
+    by some implementation of executors/strategies.
+    """
```
mikeknep, line 55: Not used anywhere?

src/gretel_trainer/benchmark/job_spec.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def make_run_key(self):
+        return RunKey((model_name(self.model), self.dataset.name))
+
+
+def model_name(model: AnyModelType) -> str:
+    if isinstance(model, GretelModel):
+        return model.name
+    else:
+        return type(model).__name__
```
mikeknep, line 65: I think this is no longer used and can be deleted; it has effectively been replaced by `_model_name` in `entrypoints.py`

mikeknep (approved):
Thanks for picking this up, sorry it ended up being more than I expected but I really love where this has ended up!

Two small final comments. Also, I tried rerunning the CI build (figuring GH Actions just had some hiccup last night) but as with the previous run attempt, the unit tests all pass but then GH Actions just hangs... 🤔 🤔

pimlock:
> Two small final comments. Also, I tried rerunning the CI build (figuring GH Actions just had some hiccup last night) but as with the previous run attempt, the unit tests all pass but then GH Actions just hangs... 🤔 🤔

Interesting... yeah, must be some kind of thread that never finishes? I will fix that.

src/gretel_trainer/benchmark/job_spec.py
pimlock: Oh yeah, this isn't used in trainer, it was something I needed in the context of test harness use case, but I did it in a different way, so this can be removed.

src/gretel_trainer/benchmark/job_spec.py
pimlock: Good point, I thought I needed these 2, as they were working on slightly different types of input and initially I thought I won't be able to put them in one place, but it was in fact possible after moving few things around.

mikeknep (approved):

PR #144, opened 2023-07-28
Remove column partitioning

mikeknep:
This has been in the backlog for a while; wanted to take a quick stab at it now because I discovered that by knocking this out we can drop the dependency on `gretel-synthetics`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 boto3~=1.20
 dask[dataframe]==2023.5.1
 gretel-client>=0.16.0
-gretel-synthetics[utils]
 jinja2~=3.1
 networkx~=3.0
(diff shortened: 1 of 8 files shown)


requirements.txt
```diff
@@ -1,6 +1,5 @@
 boto3~=1.20
 gretel-client>=0.16.0
-gretel-synthetics[utils]
```
mckornfield, line 3: 👌

mckornfield (approved):
I was going to ask about a deprecation warning and just had to scroll down 😆 
👍

mikeknep:
I ran a few manual tests to confirm that Trainer no longer performs header clustering with these changes and conditional seeding still works.

PR #145, opened 2023-07-28
Dependency tweaks

mikeknep:
- numpy is imported explicitly in the `relational.extractor` and `relational.json` modules, but had not been declared an official dependency (it was getting pulled in transitively by pandas)
- rename dev/test requirements file to be consistent with our other public Python repos (gretel-client, gretel-synthetics)
- similarly, take out the `-r requirements.txt` helper and just explicitly install both files during workflow setup
  - I believe we could get away with only installing `-r test-requirements.txt` explicitly, and then the following `pip install -e .` line would handle installing the production/runtime requirements thanks to `install_requires` in `setup.py`. I don't have a strong opinion here.
- alphabetizing is good for you

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Install deps
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements-dev.txt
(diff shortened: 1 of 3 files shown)


pimlock (approved):
> alphabetizing is good for you

😅

PR #147, opened 2023-07-31
Export helper as fixture instead of importing from test package

mikeknep:
By relying on pytest's autodetection of fixtures, we can avoid having to import `from test.relational.config`, which has been proving a bit troublesome in Bazel.

I'm not happy about the fairly gruesome diff in the `_with_json.py` test file, but trying to set the `invented_tables` fixture as `autouse=True` didn't work—it consistently printed this error:
```
Object of type "(get_invented_table_suffix: Unknown) -> dict[str, str]" is not subscriptable
```


pimlock (approved):
Nice!

PR #148, opened 2023-07-31
Misc

mikeknep:
`tests/example_config.py` was a fixture used in Benchmark v1 unit tests, no longer needed.

The pandas requirement has actually been incorrect for a while. Relational JSON uses a feature added in Pandas 1.5 (specifically, the optional `names` argument on [reset_index](https://pandas.pydata.org/pandas-docs/version/1.5/reference/api/pandas.DataFrame.reset_index.html), used by Trainer [here](https://github.com/gretelai/trainer/blob/main/src/gretel_trainer/relational/json.py#L90)). I hadn't noticed because I had 1.5 installed in my local venv.


mckornfield (approved):
definitely `misc` 😆

PR #151, opened 2023-08-02
Wrap record handler data download in smart_open

mikeknep:


mckornfield (approved):
seems good from my testing 👍

PR #153, opened 2023-08-03
Use get_artifact_handle instead of smart_open directly

mckornfield:


src/gretel_trainer/relational/sdk_extras.py
```diff
@@ -79,8 +89,30 @@ def sqs_score_from_full_report(self, report: dict[str, Any]) -> Optional[int]:
...
+        if (storage_account := os.getenv("OAUTH_STORAGE_ACCOUNT_NAME")) is not None:
+            oauth_url = "https://{}.blob.core.windows.net".format(storage_account)
+            return BlobServiceClient(
+                account_url=oauth_url, credential=DefaultAzureCredential()
+            )
+
+        if (connect_str := os.getenv("AZURE_STORAGE_CONNECTION_STRING")) is not None:
+            return BlobServiceClient.from_connection_string(connect_str)
+
+    def _get_transport_params(self, link: str) -> dict:
```
mckornfield, line 102: there are probably more places that need this, but I don't know if I want to do this preemptively or reactively...

mikeknep: I think you're probably right about more places likely needing this... Could we keep this low-level detail exclusively in the client instead of duplicating it here? I'm imagining something like:
```python
# client

class Job:
    @contextmanager
    def get_artifact_obj(self, artifact_key: str):
        link = self.get_artifact_link(artifact_key)
        transport_params = self._get_transport_params(link)
        with smart_open.open(link, "rb", transport_params=transport_params) as o:
            yield o
```

```python
# trainer

def get_record_handler_data(self, record_handler: RecordHandler) -> pd.DataFrame:
    with record_handler.get_artifact_obj("data") as data:
        pd.read_csv(data)
``` 
(very open to something other than "obj", but you get the idea)

mckornfield: yeah I can dig it, `get_artifact_handle` ?

PR #154, opened 2023-08-10
Handle nested lists of nested objects

mikeknep:
Fixes a bug where relational JSON could not recompose lists of nested objects. See unit test for example. I've also filed a backlog ticket to improve how we handle the edge case documented in the second unit test.

This diff is pretty noisy because I wanted to add some comments / refactors / naming improvements to help explain what's going on and reduce the amount of time spent just figuring out what's going on next time we have to visit this. The actual fix was shockingly trivial, I will call it out inline.

### DIFF
--- a/src/gretel_trainer/relational/json.py
+++ b/src/gretel_trainer/relational/json.py
@@ -314,30 +314,48 @@ def _denormalize_json(
     table_name_mappings: dict[str, str],
     original_table_name: str,
 ) -> pd.DataFrame:
-    table_names = list(table_name_mappings.values())
+    # Throughout this function, "provenance name" refers to the json-breadcrumb-style name (e.g. foo^bar>baz)
+    # and "node name" refers to the table name as it appears in the RelationalData graph (e.g. foo_invented_{uuid})
+
+    # The provided `table_name_mappings` argument (from producer metadata) maps from provenance name to node name.
+    # `inverse_table_name_mappings` inverts that mapping (so: node name keys, provenance name values).
+    # `table_dict` replaces the keys (node namees) in `tables` with corresponding provenance names
+    table_node_names = list(table_name_mappings.values())
     inverse_table_name_mappings = {v: k for k, v in table_name_mappings.items()}
-    table_dict: dict = {inverse_table_name_mappings[k]: v for k, v in tables.items()}
-    for table_name in list(reversed(table_names)):
+    table_dict = {inverse_table_name_mappings[k]: v for k, v in tables.items()}
+
+    get_table = lambda node_name: _get_table(
+        rel_data, inverse_table_name_mappings, table_dict, node_name
+    )
+
+    for table_name in list(reversed(table_node_names)):
+        table_df = get_table(table_name)
         table_provenance_name = inverse_table_name_mappings[table_name]
-        empty_fallback = pd.DataFrame(
-            data={col: [] for col in rel_data.get_table_columns(table_name)},
-        )
-        table_df = table_dict.get(table_provenance_name, empty_fallback)
 
         if table_df.empty and _is_invented_child_table(table_name, rel_data):
-            p_name = rel_data.get_foreign_keys(table_name)[0].parent_table_name
-            parent_name = inverse_table_name_mappings[p_name]
+            # Mutate the parent dataframe by adding a column with empty lists
+
+            # We know invented child tables have exactly one foreign key
+            fk = rel_data.get_foreign_keys(table_name)[0]
+            parent_node_name = fk.parent_table_name
+            parent_provenance_name = inverse_table_name_mappings[parent_node_name]
             col_name = get_parent_column_name_from_child_table_name(
                 table_provenance_name
             )
-            kwargs = {col_name: table_dict[parent_name].apply(lambda x: [], axis=1)}
-            table_dict[parent_name] = table_dict[parent_name].assign(**kwargs)
+            parent_df = get_table(parent_node_name)
+            kwargs = {col_name: parent_df.apply(lambda x: [], axis=1)}
+            # Here's where we mutate the DF
+            table_dict[parent_provenance_name] = parent_df.assign(**kwargs)
         else:
+            # First, make a version of `table_df` with column names altered to send to `unflatten`
+            # See: https://github.com/dairiki/unflatten/#synopsis
             col_names = [col for col in table_df.columns if FIELD_SEPARATOR in col]
             new_col_names = [col.replace(FIELD_SEPARATOR, ".") for col in col_names]
             flat_df = table_df[col_names].rename(
                 columns=dict(zip(col_names, new_col_names))
             )
+
+            # Turn this dataframe into a dict to send to `unflatten`...
             flat_dict = {
                 k: {
                     k1: v1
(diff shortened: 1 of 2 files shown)


src/gretel_trainer/relational/json.py
```diff
@@ -346,19 +364,24 @@ def _denormalize_json(
...
                 pd.DataFrame.from_dict(
                     {k: unflatten(v) for k, v in flat_dict.items()}, orient="index"
                 )
             )
-            nested_df = table_df.join(dict_df).drop(columns=col_names)
+
+            # Join the flattened dict_df onto table_df.
+            # You might think we could drop the original non-nested columns at this point,
+            # but under certain JSON shapes (e.g. nested lists of objects) we still need them present.
+            nested_df = table_df.join(dict_df)
```
mikeknep, line 377: This is the only change we really had to make to fix the bug. Github is mucking up the diff a bit—compare it to the deleted line 354: we were dropping columns too aggressively, leading to `KeyError`s when performing `groupby` calls a few lines lower.

src/gretel_trainer/relational/json.py
```diff
@@ -314,30 +314,48 @@ def _denormalize_json(
     table_name_mappings: dict[str, str],
     original_table_name: str,
 ) -> pd.DataFrame:
-    table_names = list(table_name_mappings.values())
+    # Throughout this function, "provenance name" refers to the json-breadcrumb-style name (e.g. foo^bar>baz)
```
mckornfield, line 317: should this be a function docstring, or is that not appropriate since it also talks about how things are used below?

src/gretel_trainer/relational/json.py
```diff
@@ -314,30 +314,48 @@ def _denormalize_json(
...
+    # and "node name" refers to the table name as it appears in the RelationalData graph (e.g. foo_invented_{uuid})
+
+    # The provided `table_name_mappings` argument (from producer metadata) maps from provenance name to node name.
+    # `inverse_table_name_mappings` inverts that mapping (so: node name keys, provenance name values).
+    # `table_dict` replaces the keys (node namees) in `tables` with corresponding provenance names
```
mckornfield, line 322:
```suggestion
    # `table_dict` replaces the keys (node names) in `tables` with corresponding provenance names
```

src/gretel_trainer/relational/json.py
```diff
@@ -367,16 +390,32 @@ def _denormalize_json(
                 col_name = get_parent_column_name_from_child_table_name(
                     table_provenance_name
                 )
-                table_dict[parent_name] = table_dict[parent_name].join(
```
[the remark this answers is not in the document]
mckornfield: ah I like the fallback, maybe `get_table_or_empty_df` would be clearer but 🤷 maybe too wordy

src/gretel_trainer/relational/json.py
```diff
@@ -367,16 +390,32 @@ def _denormalize_json(
...
+    table_dict: dict[str, pd.DataFrame],
+    node_name: str,
+) -> pd.DataFrame:
+    """
+    Helper function used inside _denormalize_json to safely retrieve a table dataframe.
+    If `node_name` is not present in `table_dict` (e.g. because the Gretel Job failed),
+    return an empty dataframe with the expected columns.
+    """
+    empty_fallback = pd.DataFrame(
+        data={col: [] for col in rel_data.get_table_columns(node_name)},
```
mckornfield, line 413: nit, could also do `pd.DataFrame(columns=rel_data.get_table_columns(node_name))`
```
Pd.>>> pd.DataFrame(columns=["a","b","c"])
Empty DataFrame
Columns: [a, b, c]
Index: []
>>> pd.DataFrame(data={col:[] for col in ["a","b","c"]})
Empty DataFrame
Columns: [a, b, c]
Index: []
```

tests/relational/test_relational_data_with_json.py
```diff
@@ -713,6 +713,150 @@ def test_restore_with_empty_tables(bball, invented_tables):
     assert jimmy["suspensions"] == []
 
 
+@pytest.fixture
+def nested_listed_objects(tmpdir):
+    json = """
+{
+    "Records": [
```
mckornfield, line 720: hehe nice formatting

tests/relational/test_relational_data_with_json.py
```diff
@@ -713,6 +713,150 @@ def test_restore_with_empty_tables(bball, invented_tables):
...
+
+    pdtest.assert_frame_equal(
+        restored["demo"],
+        pd.DataFrame(
+            data={
+                "Records": [
+                    {
+                        "userAgent": "abc",
+                        "responseElements": {
+                            "accountAttributes": [
```
mckornfield, line 843: RIP 😢

tests/relational/test_relational_data_with_json.py
```diff
@@ -713,6 +713,150 @@ def test_restore_with_empty_tables(bball, invented_tables):
...
+"""
+    json_df = pd.read_json(json)
+
+    rel_data = RelationalData(directory=tmpdir)
+    rel_data.add_table(name="demo", primary_key=None, data=json_df)
+
+    return rel_data
+
+
+def test_nested_listed_objects(nested_listed_objects):
```
mckornfield, line 747: `
```suggestion
def test_nested_list_objects(nested_listed_objects):
```

src/gretel_trainer/relational/json.py
```diff
@@ -314,30 +314,48 @@ def _denormalize_json(
...
+            fk = rel_data.get_foreign_keys(table_name)[0]
+            parent_node_name = fk.parent_table_name
+            parent_provenance_name = inverse_table_name_mappings[parent_node_name]
             col_name = get_parent_column_name_from_child_table_name(
                 table_provenance_name
             )
-            kwargs = {col_name: table_dict[parent_name].apply(lambda x: [], axis=1)}
-            table_dict[parent_name] = table_dict[parent_name].assign(**kwargs)
+            parent_df = get_table(parent_node_name)
+            kwargs = {col_name: parent_df.apply(lambda x: [], axis=1)}
```
mckornfield, line 346: what does this do? makes an empty list for every row?

mckornfield (approved):
some questions/ nitty comments

benmccown (approved):
Interested in some of the answers to Matt's questions, but otherwise LGTM.

src/gretel_trainer/relational/json.py
pimlock: +1 to Matt's question and I think that `apply` and `assign` combo makes it a bit hard to understand.

Maybe writing it in this way would be easier to grasp?
```python
parent_df[col_name] = [[] for _ in range(len(parent_df))]
```

I think modifying `parent_df` is okay here, but if we didn't want to do that, we could
```python
parent_df.assign(**{col_name: [[] for _ in range(len(parent_df))]})
```

pimlock (approved):
LGTM with one suggestion for potential simplification.

gracecvking (approved):

tests/relational/test_relational_data_with_json.py
mikeknep: `list_objects` feels just slightly ambiguous to me, I'm going with `nested_lists_of_objects`

src/gretel_trainer/relational/json.py
mikeknep: Yep, yours is definitely better

src/gretel_trainer/relational/json.py
mikeknep: Yeah this is definitely more clear, thanks. BTW, we _need_ to modify `parent_df` so that the column we add is present on subsequent iterations through the tables loop.

src/gretel_trainer/relational/json.py
mikeknep: (Disclaimer: I'm not as consistent about comments and/or docstrings as I ought to be.)

This feels more like a note to developers who need to grok the implementation for debugging/modifying/refactoring, as opposed to a note to clients of the function (who won't know or need to care about "provenance" vs "node"), so to me this makes more sense as a comment than a docstring.

